PR #4, opened 2018-01-22
remove shuttle updates before the vehicle starts on the route

paulswartz:
asana: https://app.asana.com/0/505721188639414/510039796888298

The rationale for this is to avoid making incorrect predictions for vehicles on the opposite side of a shuttle:
```
A <-> B <shuttle> C <-> D
```

For many prediction systems, a train on an `A -> D` trip will have predictions made for a return `D -> A` trip on the same block. However, since we know the train won't move past the shuttle, we want to remove the `D -> A` trip predictions.


paulswartz:
Talked with @dmaltzan and turns out this isn't needed.

PR #40, opened 2018-03-14
pull grouping/group filtering into the MergeFilter

paulswartz:
Since there's no parallelization happening here, we can pull the grouping and
group filtering into the same ProducerConsumer to save some copying.


paulswartz:
The missing coverage is due to to the `debug` logs, which aren't hit during testing.

PR #45, opened 2018-04-04
fix for TripUpdate w/o StopTimeUpdates

paulswartz:
Instead of ignoring them, we'll keep them but only if their status isn't
`SCHEDULED`.


paulswartz:
The data we were getting from Swiftly looked like this:
```
trip_update {
  trip {
    trip_id "whatever"
    schedule_relationship CANCELED
  }
  ... no stop_time_update ...
}
```

Before, we expected that we'd never see a TripUpdate without any StopTimeUpdate values: what's the point of including it? And so, when parsing, we'd crash.

This PR changes the parser to include the TripUpdate even if there aren't any StopTimeUpdates. We still won't output a TripUpdate without StopTimeUpdates, unless the trip is listed as something other than SCHEDULED: this would, in the future, allow us to propagate a cancelled trip up to our clients, even without the individual stops being SKIPPED in the output.

PR #56, opened 2018-06-29
better handling of shuttles at the start of trips

paulswartz:
Previously, the behavior for shuttles at the start of trips was to cancel the
entire trip. However, this isn't the desired behavior: trips skipped at the
beginning are generally running the second part of the trip.

The new behavior is to only skip the stop times at the beginning. If boarding
is available at the last stop in the shuttle, then that stop will not be
skipped but instead only lose its `arrival_time`.

### DIFF
--- a/test/concentrate/group_filter/shuttle_test.exs
+++ b/test/concentrate/group_filter/shuttle_test.exs
@@ -97,7 +97,7 @@ defmodule Concentrate.GroupFilter.ShuttleTest do
       assert StopTimeUpdate.schedule_relationship(after_shuttle) == :SKIPPED
     end
 
-    test "everything is skipped if the first stop is shuttled" do
+    test "only the first part is skipped if the first stop is shuttled" do
       group =
         {TripUpdate.new(
            trip_id: @trip_id,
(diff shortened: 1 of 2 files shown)


test/concentrate/group_filter/shuttle_test.exs
```diff
@@ -112,22 +112,58 @@ defmodule Concentrate.GroupFilter.ShuttleTest do
...
+             departure_time: @valid_date_time
+           )
+         ]}
+
+      {_tu, [], reduced} = filter(group, @module)
+      assert [start, stop, after_shuttle] = reduced
+      assert StopTimeUpdate.schedule_relationship(start) == :SKIPPED
+      assert StopTimeUpdate.schedule_relationship(stop) == :SCHEDULED
+      assert StopTimeUpdate.arrival_time(stop) == nil
+      assert StopTimeUpdate.schedule_relationship(after_shuttle) == :SCHEDULED
```
paulswartz: @dmaltzan here's the test for adjusting the arrival time.

The predictions originally look something like this:

| stop_id | arrival_time | departure_time | schedule_relationship |
| - | - | - | - |
| shuttle_1 | 1 | 2 | `SCHEDULED` |
| shuttle_stop | 3 | 4 | `SCHEDULED` |
| after_shuttle | 5 | 6 | `SCHEDULED` |

For the `shuttle_stop` stop, where you're allowed to board, we want to show that the train departs from the stop, but doesn't have a previous stop. We end with data like this:

| stop_id | arrival_time | departure_time | schedule_relationship |
| - | - | - | - |
| shuttle_1 | - | - | `SKIPPED` |
| shuttle_stop | - | 4 | `SCHEDULED` |
| after_shuttle | 5 | 6 | `SCHEDULED` |

PR #83, opened 2019-04-01
exponential backoff for InstanceMeta

paulswartz:


paulswartz:
We've worked around this with a Docker HEALTHCHECK.

PR #98, opened 2019-09-17
Include route_pattern_id in TripUpdate

handorff:
If `route_pattern_id` are present in `TripUpdate`, parse them and include them in enhanced output.

### DIFF
--- a/lib/concentrate/encoder/trip_updates_enhanced.ex
+++ b/lib/concentrate/encoder/trip_updates_enhanced.ex
@@ -3,17 +3,21 @@ defmodule Concentrate.Encoder.TripUpdatesEnhanced do
   Encodes a list of parsed data into an enhanced.pb file.
   """
   @behaviour Concentrate.Encoder
-  alias Concentrate.StopTimeUpdate
+  alias Concentrate.{StopTimeUpdate, TripUpdate}
   import Concentrate.Encoder.GTFSRealtimeHelpers
 
   @impl Concentrate.Encoder
   def encode_groups(groups) when is_list(groups) do
     message = %{
       header: feed_header(),
-      entity: trip_update_feed_entity(groups, &build_stop_time_update/1, true)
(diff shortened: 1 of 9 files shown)


lib/concentrate/parser/gtfs_realtime.ex
```diff
@@ -122,6 +122,7 @@ defmodule Concentrate.Parser.GTFSRealtime do
       TripUpdate.new(
         trip_id: Map.get(trip, :trip_id),
         route_id: Map.get(trip, :route_id),
+        route_pattern_id: Map.get(trip, :route_pattern_id),
```
paulswartz: This will always be `nil` since it's not part of the protobuf specification: you can leave it out.

lib/concentrate/encoder/trip_updates_enhanced.ex
```diff
@@ -10,7 +10,7 @@ defmodule Concentrate.Encoder.TripUpdatesEnhanced do
   def encode_groups(groups) when is_list(groups) do
     message = %{
       header: feed_header(),
-      entity: trip_update_feed_entity(groups, &build_stop_time_update/1)
+      entity: trip_update_feed_entity(groups, &build_stop_time_update/1, true)
```
paulswartz: I don't love passing in the boolean; can we pass in a function instead which adds the relevant data to the `TripDescriptor`?

test/concentrate/encoder/trip_updates_enhanced_test.exs
```diff
@@ -85,5 +85,28 @@ defmodule Concentrate.Encoder.TripUpdatesEnhancedTest do
...
+            "trip_update" => %{
+              "trip" => %{
+                "route_pattern_id" => rp_id
+              }
+            }
+          }
+        ]
+      } = encoded
+
+      assert rp_id == "pattern"
```
paulswartz: You can put `"pattern"` into the map on line 102 and assert the whole thing.

PR #99, opened 2019-10-11
Support referencing system environment variables in headers

handorff:


test/concentrate_test.exs
```diff
@@ -146,5 +146,45 @@ defmodule ConcentrateTest do
...
+      System.put_env(env_var, env_var_value)
+
+      try do
+        config = parse_json_configuration(body)
+        assert config[:sources][:gtfs_realtime][:name_1] == "url_1"
+
+        assert config[:sources][:gtfs_realtime][:name_2] ==
+                 {"url_2", [headers: %{"Authorization" => "secret_key"}]}
+
+        assert config[:sources][:gtfs_realtime][:name_3] ==
```
paulswartz: You really only need this assertion, and the single source in the JSON. It'll be easier in the future to understand which parts are important (only that  one header).

PR #100, opened 2019-12-11
fix: filter out empty StopTimeUpdate records

paulswartz:
For StopTimeUpdate records in the Protobuf export, we don't include the
`status` field. This can result in records like this:

```
entity {
  id: "CR-Weekday-Fall-19-616"
  trip_update {
    trip {
      trip_id: "CR-Weekday-Fall-19-616"
      start_date: "20191210"
      route_id: "CR-Needham"
      direction_id: 1
    }
    stop_time_update {
      stop_sequence: 11
      stop_id: "South Station"
    }
  }
}

```

That StopTimeUpdate does not have any useful information, so instead we drop
it from the rendered version. This had some knock-on effects on test cases,
where we weren't providing any data for StopTimeUpdates we expected to be in
the output.

### DIFF
--- a/lib/concentrate/encoder/gtfs_realtime_helpers.ex
+++ b/lib/concentrate/encoder/gtfs_realtime_helpers.ex
@@ -198,7 +198,7 @@ defmodule Concentrate.Encoder.GTFSRealtimeHelpers do
 
     stop_time_update =
       case stus do
-        [_ | _] -> Enum.map(stus, stop_time_update_fn)
+        [_ | _] -> render_stop_time_updates(stus, stop_time_update_fn)
         [] -> nil
       end
 
@@ -231,6 +231,18 @@ defmodule Concentrate.Encoder.GTFSRealtimeHelpers do
 
   defp build_trip_update_entity(_, _, _) do
     []
+  end
+
+  defp render_stop_time_updates(stus, stop_time_update_fn) do
+    Enum.flat_map(stus, fn stu ->
+      case stop_time_update_fn.(stu) do
+        :skip ->
+          []
+
+        update ->
+          [update]
+      end
+    end)
   end
 
   defp trip_update_vehicle(_update, [vp | _]) do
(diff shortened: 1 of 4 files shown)


lib/concentrate/encoder/gtfs_realtime_helpers.ex
```diff
@@ -225,6 +225,21 @@ defmodule Concentrate.Encoder.GTFSRealtimeHelpers do
     []
   end
 
+  defp render_stop_time_updates(stus, stop_time_update_fn) do
+    :lists.filtermap(
```
losvedir: Out of curiosity, is there a benefit to `:lists.filtermap/1` over `Enum.flat_map` and returning `[]` or `[elem]`, which was what I had thought was the sort of conventional technique for this sort of thing?

paulswartz: I assumed it was more efficient (since it has fewer cases to handle), but it's about the same. I'll switch to `Enum` since that's more obvious to people.

test/concentrate/encoder/trip_updates_test.exs
```diff
@@ -79,6 +81,19 @@ defmodule Concentrate.Encoder.TripUpdatesTest do
              } = decoded
     end
 
+    test "stop time updates with only a boarding status are removed" do
+      initial = [
+        TripUpdate.new(trip_id: "1"),
+        StopTimeUpdate.new(trip_id: "1", status: "status"),
+        StopTimeUpdate.new(trip_id: "1", arrival_time: 1)
+      ]
```
losvedir: Maybe it would also be good to include a `StopTimeUpdate` that has both a `:status` and more information, just to protect against the case that the logic simply filters out StopTimeUpdates that contain statuses? Since at a quick glance I don't see any other test cases in this file that have a `:status`.

paulswartz: can do!

PR #134, opened 2020-04-15
feat: pass occupancy_status/occupancy_percentage through

paulswartz:
We don't currently have these fields in production, but we're looking to
integrate them in the short term.

### DIFF
--- a/lib/concentrate/encoder/vehicle_positions.ex
+++ b/lib/concentrate/encoder/vehicle_positions.ex
@@ -74,7 +74,9 @@ defmodule Concentrate.Encoder.VehiclePositions do
       stop_id: VehiclePosition.stop_id(vp),
       current_stop_sequence: VehiclePosition.stop_sequence(vp),
       current_status: VehiclePosition.status(vp),
-      timestamp: VehiclePosition.last_updated(vp)
+      timestamp: VehiclePosition.last_updated(vp),
+      occupancy_status: VehiclePosition.occupancy_status(vp),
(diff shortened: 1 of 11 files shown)


lemald:
Is my understanding correct that `gtfs_realtime_proto.erl` is auto-generated from the corresponding `.proto` file? That's what the README in the `src` directory would suggest, but I just wanted to double-check.

paulswartz:
> Is my understanding correct that `gtfs_realtime_proto.erl` is auto-generated from the corresponding `.proto` file? That's what the README in the `src` directory would suggest, but I just wanted to double-check.

That's correct; the .proto file was updated from the latest `master` branch which includes the new `occupancy_percentage` field.

lib/concentrate/encoder/vehicle_positions.ex
```diff
@@ -74,7 +74,9 @@ defmodule Concentrate.Encoder.VehiclePositions do
       stop_id: VehiclePosition.stop_id(vp),
       current_stop_sequence: VehiclePosition.stop_sequence(vp),
       current_status: VehiclePosition.status(vp),
-      timestamp: VehiclePosition.last_updated(vp)
+      timestamp: VehiclePosition.last_updated(vp),
+      occupancy_status: VehiclePosition.occupancy_status(vp),
+      occupancy_percentage: VehiclePosition.occupancy_percentage(vp)
```
lemald, line 79: I see tests for the corresponding fields in `vehicle_positions_enhanced`, but not for the `vehicle_positions` encoder.

PR #137, opened 2020-04-27
Log vehicle timestamp mismatch

lemald:
Asana ticket: [🚝 log vehicle timestamps which are greater than the reported feed timestamp](https://app.asana.com/0/584764604969369/1168026177896885/f)

### DIFF
--- a/lib/concentrate/parser/gtfs_realtime.ex
+++ b/lib/concentrate/parser/gtfs_realtime.ex
@@ -4,6 +4,7 @@ defmodule Concentrate.Parser.GTFSRealtime do
   """
   @behaviour Concentrate.Parser
   alias Concentrate.Parser.Helpers
+  require Logger
 
   alias Concentrate.{Alert, Alert.InformedEntity, StopTimeUpdate, TripUpdate, VehiclePosition}
   @impl Concentrate.Parser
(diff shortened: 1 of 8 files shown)


lib/concentrate/producer/httpoison.ex
```diff
@@ -27,10 +27,10 @@ defmodule Concentrate.Producer.HTTPoison do
     parser =
       case Keyword.fetch!(opts, :parser) do
         module when is_atom(module) ->
-          &module.parse(&1, [])
+          &module.parse(&1, feed_url: url)
 
         {module, opts} when is_atom(module) and is_list(opts) ->
-          &module.parse(&1, opts)
+          &module.parse(&1, opts ++ [feed_url: url])
```
paulswartz, line 33: Better the other way: `[feed_url: url] ++ opts` since it doesn't require iterating over `opts` to build a new list.

lib/concentrate/parser/gtfs_realtime.ex
```diff
@@ -42,11 +53,21 @@ defmodule Concentrate.Parser.GTFSRealtime do
 
       vehicle = vp.vehicle
       position = vp.position
+      id = Map.get(vehicle, :id)
+      timestamp = Map.get(vp, :timestamp)
+
+      if is_integer(timestamp) && timestamp > feed_timestamp do
```
paulswartz, line 59: Should you also be checking that `feed_timestamp` is an integer here?

lib/concentrate/producer/httpoison.ex
lemald: Ah, whoops, and usually that's an optimization that I'm actually picky about.

lib/concentrate/parser/gtfs_realtime.ex
lemald: Yeah, probably.

PR #174, opened 2021-02-03
build: don't attempt to install awscli via pip

lemald:
Similar to what I did with the parking tweets repo yesterday.


digitalcora (approved):
🚫 🐍 2️⃣

PR #183, opened 2021-04-09
ci: add Github action for CI

paulswartz:
Asana: https://app.asana.com/0/search/1200172368489010/1199946732081250

This does not yet handle the deployment bit, but does run the Elixir tests (including coverage), Credo, and Dialyzer.

You can see a full example in my forked PR: https://github.com/paulswartz/concentrate/pull/1


.github/workflows/elixir.yml
```diff
@@ -0,0 +1,49 @@
...
+        key: ${{ runner.os }}-mix-${{ hashFiles('**/mix.lock') }}
+        restore-keys: ${{ runner.os }}-mix-
+    - name: Install dependencies
+      run: mix deps.get
+    - name: Check formatting
+      run: mix format --check-formatted
+    - name: Credo
+      run: mix credo --strict
+    - name: Run tests
+      run: env MIX_ENV=test mix do compile --force --warnings-as-errors, test --cover
```
digitalcora, line 33: Maybe this could be two separate steps, `env MIX_ENV=test mix compile ...` and `mix test --cover`? Then if the compile fails, it doesn't look like the tests failed (because "Run tests" is the step that failed).

digitalcora (commented):
Sorry this took a second, had to go read a bit about GitHub Actions (never used it before!).

Really unfortunate there's apparently no way to have multiple parallel jobs share a sequence of "setup" steps — this is so common among other CI services I've used that I'm surprised it wasn't considered an essential feature here. You might have found it already but [this seems like the issue to track](https://github.com/actions/runner/issues/646) for alternatives and when it might be available officially (it looks like there's a way to fake it using the matrix feature?).

.github/workflows/elixir.yml
paulswartz: Good idea, done in latest commit.

digitalcora:
Hm... not sure the coverage report is all that useful in its current form, since it doesn't show coverage _changes_. Without that, I'm not sure what we can practically do with the information provided; it seems like it will just become noise that we don't pay attention to.

I'm looking through the GitHub Actions Marketplace™... have you seen [this LCOV report action](https://github.com/marketplace/actions/report-lcov)? It seems to support showing the coverage of changed files.

paulswartz:
> I'm looking through the GitHub Actions Marketplace™... have you seen [this LCOV report action](https://github.com/marketplace/actions/report-lcov)? It seems to support showing the coverage of changed files.

It doesn't appear to actually post a comment. I see the uploaded artifact [here](https://github.com/paulswartz/concentrate/actions/runs/741958885) but no comment in the pull request.

digitalcora:
Maybe it doesn't make the comment unless the job is `on: pull_request` as in the action's docs. I'm not sure if this would still allow it to run coverage on each commit and update/redo the comment; if not, maybe this particular action doesn't work for us.

paulswartz:
Yeah, adding `pull_request` to the jobs seems to work okay. I'll see what happens with the comment if I rebase the top commit.

digitalcora:
At a higher level I'm not sure how to "review" the part of this PR where, in addition to migrating to GitHub Actions (which is mentioned in the Asana task), we are migrating our coverage reporting away from Codecov (which isn't). If that is a goal, can we either add acceptance criteria for it or split it into a separate task?

paulswartz:
I'll add it to the AC: see https://mbta.slack.com/archives/GP5QJM98A/p1612968477036400 for the details on the switch.

digitalcora:
I was mostly interested in acceptance criteria for the sake of evaluating the code coverage solution more "objectively" than just my own viewpoint, which was that it didn't provide any useful information. Useful to me would be: "I know if my PR introduced any uncovered code, and if so, where it is". What do you think?

paulswartz:
It's unlikely that we'll get an out-of-the-box solution which replicates 💯 of what we got with CodeCov. The second uploader posts a comment with the coverage for the changed files, which seems like a good place to start.

digitalcora:
> The second uploader posts a comment with the coverage for the changed files, which seems like a good place to start.

Has that happened on this PR? I thought you were still trying to get that one to work. The ["total coverage"](https://github.com/mbta/concentrate/pull/183#commitcomment-49416494) bot was the one I didn't think was useful.

paulswartz:
> > The second uploader posts a comment with the coverage for the changed files, which seems like a good place to start.
> 
> Has that happened on this PR? I thought you were still trying to get that one to work. The ["total coverage"](https://github.com/mbta/concentrate/pull/183#commitcomment-49416494) bot was the one I didn't think was useful.

https://github.com/paulswartz/concentrate/pull/1#issuecomment-817983437

digitalcora (approved):

PR #191, opened 2021-04-21
ci: workflow to deploy (dev/dev-green/prod)

paulswartz:
required secrets in GitHub:
- DOCKER_REPO
- AWS_ACCESS_KEY_ID
- AWS_SECRET_ACCESS_KEY

### DIFF
--- a/.github/actions/build-push/action.yml
+++ b/.github/actions/build-push/action.yml
@@ -1,0 +1,35 @@
+name: 'Build/push'
+description: 'Build a Docker image and push to ECR'
+inputs:
+  aws-access-key-id:
+    description: "AWS_ACCESS_KEY_ID"
+    required: true
+  aws-secret-access-key:
+    description: "AWS_SECRET_ACCESS_KEY"
+    required: true
+  aws-region:
+    description: "AWS region to use"
+    required: true
+    default: "us-east-1"
+  docker-repo:
+    description: "ECR Docker repo to push to"
+    required: true
+outputs:
+  docker-tag:
+    description: "Docker Tag"
+    value: ${{ steps.docker.outputs.tag }}
+runs:
+  using: "composite"
+  steps:
+    - run: echo "::set-output name=tag::${{ inputs.docker-repo }}:git-$(git rev-parse --short HEAD)"
+      id: docker
+      shell: bash
+    - run: docker build --pull -t ${{ steps.docker.outputs.tag }} .
+      shell: bash
+    - run: aws ecr get-login-password --region ${{ inputs.aws-region }} | docker login --username AWS --password-stdin ${{ inputs.docker-repo }}
+      env:
+        AWS_ACCESS_KEY_ID: ${{ inputs.aws-access-key-id }}
+        AWS_SECRET_ACCESS_KEY: ${{ inputs.aws-secret-access-key }}
+      shell: bash
+    - run: docker push ${{ steps.docker.outputs.tag }}
+      shell: bash
--- a/.github/actions/build-push/action.yml
+++ b/.github/actions/build-push/action.yml
@@ -1,32 +1,34 @@
-name: 'Build/push'
-description: 'Build a Docker image and push to ECR'
+name: Build/push
+description: Build a Docker image and push to ECR
 inputs:
   aws-access-key-id:
-    description: "AWS_ACCESS_KEY_ID"
+    description: AWS_ACCESS_KEY_ID
     required: true
   aws-secret-access-key:
-    description: "AWS_SECRET_ACCESS_KEY"
+    description: AWS_SECRET_ACCESS_KEY
     required: true
   aws-region:
-    description: "AWS region to use"
+    description: AWS region to use
     required: true
-    default: "us-east-1"
+    default: us-east-1
   docker-repo:
-    description: "ECR Docker repo to push to"
+    description: ECR Docker repo to push to
     required: true
 outputs:
   docker-tag:
-    description: "Docker Tag"
+    description: Docker Tag
     value: ${{ steps.docker.outputs.tag }}
 runs:
-  using: "composite"
+  using: composite
   steps:
     - run: echo "::set-output name=tag::${{ inputs.docker-repo }}:git-$(git rev-parse --short HEAD)"
       id: docker
       shell: bash
     - run: docker build --pull -t ${{ steps.docker.outputs.tag }} .
       shell: bash
-    - run: aws ecr get-login-password --region ${{ inputs.aws-region }} | docker login --username AWS --password-stdin ${{ inputs.docker-repo }}
+    - run: >
+        aws ecr get-login-password --region ${{ inputs.aws-region }}
+        | docker login --username AWS --password-stdin ${{ inputs.docker-repo }}
       env:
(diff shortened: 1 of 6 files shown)


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,35 @@
+name: Elixir CI (pull request)
+
+on: [pull_request_target]
```
digitalcora, line 3: The docs say this about [`pull_request_target`](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target):

> This event runs in the context of the base of the pull request, rather than in the merge commit

I'm not 100% sure what "runs in the context of" means here, but does this mean the checked out code will be the PR base, and so the reported coverage will always be that of the main branch? [This blog post](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) seems to describe a more complex workflow that needs to be followed if the situation requires _"building the untrusted code and also [needing] write permissions to update the PR with e.g. code coverage results or other test results"_. But then, I'm not sure how the `zgosalvez` action is intended to work, since the docs also say the `GITHUB_TOKEN` provided in the context of a `pull_request` action (which is what their README uses) only has read permissions.

digitalcora (commented):
Learning more and more about Actions :octocat:

.github/actions/build-push/action.yml
```diff
@@ -0,0 +1,35 @@
...
+outputs:
+  docker-tag:
+    description: "Docker Tag"
+    value: ${{ steps.docker.outputs.tag }}
+runs:
+  using: "composite"
+  steps:
+    - run: echo "::set-output name=tag::${{ inputs.docker-repo }}:git-$(git rev-parse --short HEAD)"
+      id: docker
+      shell: bash
```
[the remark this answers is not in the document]
paulswartz: Not for [composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action): https://github.com/paulswartz/concentrate/actions/runs/771318458

.github/actions/build-push/action.yml
```diff
@@ -0,0 +1,35 @@
+name: 'Build/push'
+description: 'Build a Docker image and push to ECR'
```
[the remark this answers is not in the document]
paulswartz: 👍 I'll see if there's a formatter I can use.

.github/actions/deploy/action.yml
```diff
@@ -0,0 +1,34 @@
...
+    description: "ECS service to deploy"
+    required: true
+  docker-tag:
+    description: "Docker tag to deploy"
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - run: bash ${{ github.action_path }}/deploy.sh
+      shell: bash
```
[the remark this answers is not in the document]
paulswartz: I'll see, but I didn't mark the shell script as executable in Git so I'm not sure it'll work directly.

.github/actions/deploy/deploy.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+set -e -u
```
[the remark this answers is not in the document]
paulswartz: Different environment variables, and once we're off Semaphore I'll delete the `semaphore` directory entirely.

.github/workflows/dev-green.yml
```diff
@@ -0,0 +1,28 @@
+name: Deploy to Dev-green
+
+on:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
paulswartz: Thanks, this is leftover from when `branches` was in here as well.

.github/workflows/pull_request.yml
paulswartz: It means that it runs the action from the `master` branch, not from my fork. This prevents external people from forking the repo to change the Actions to expose your secrets, and then having the action run automatically. Instead, this runs the `pull_request` action as defined on the branch that's being merged into (`master` here) which we, not the attacker, control, and so it can be given access to a `GITHUB_TOKEN` with permission to create a comment.

paulswartz: You can see the comparison in this PR: the "Build and test (pull_request_target)" is running the broken Dialyzer step from `master`, while "Build and test (pull request)" is running the fixed version on this branch.

.github/actions/build-push/action.yml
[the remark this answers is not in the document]
digitalcora: Well that's annoying. But I guess if it's required, it's required!

.github/actions/deploy/action.yml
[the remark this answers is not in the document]
digitalcora: I'd be for making the script `+x`. Is there much point to it having a `#!` otherwise?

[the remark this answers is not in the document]
paulswartz: https://github.com/paulswartz/concentrate/runs/2404284164?check_suite_focus=true looks like it's working without the extra `bash`.

.github/workflows/pull_request.yml
digitalcora: Right, I understand the security things, but the docs and that section of the blog post suggested to me that the code which is checked out in a `pull_request_target` context (by the `checkout` action) would be the code from the PR base — in which case it wouldn't be able to report on the code coverage of changes in the PR. In particular it sounds like we are in that exact "building the untrusted code and also needing write permissions" scenario described in the blog post. Does this work as-is within your fork? (correctly showing coverage changes)

paulswartz: Oh, it's not because upstream is only looking for `pull_request` not `pull_request_target`. I'll fork it and update.

paulswartz: But that's not going to work until the PR lands, since it'll continue to use the version from `mbta` :)

digitalcora: > It means that it runs the action from the `master` branch, not from my fork.

Right, I understand that part, but the docs and that blog post suggested the commit which is checked out in this context (by the `checkout` action) is also from the `master` branch — meaning the coverage reporter would just be reporting the coverage of `master` and not the PR. In particular it sounds like we're in that exact "building the untrusted code and also needing write permissions" scenario the blog post describes. Can we confirm the commit we're getting within this workflow is actually the head of the PR?

paulswartz: - This PR has two separate workflows (elixir.yml and pull_request.yml) now.
- From the pull_request_target build, you can see here that's it's checking out the correct SHA https://github.com/mbta/concentrate/pull/191/checks?check_run_id=2404383297#step:2:366

digitalcora: That's the correct SHA? It's `f0ee228`, which is the commit currently on the head of `master`, not the head of this branch.

paulswartz: Oh, I see what you're saying. I'll look into it some more.

paulswartz: Given that this is likely going to be more complicated, I've pulled all the CI changes out of this PR: only the deployment changes are up for review now.

paulswartz:
Replaced by #192.

PR #192, opened 2021-04-22
Deploy using GitHub Actions

paulswartz:
required secrets in GitHub:
- DOCKER_REPO
- AWS_ACCESS_KEY_ID
- AWS_SECRET_ACCESS_KEY

Replaces #191 .

### DIFF
--- a/.github/actions/build-push/action.yml
+++ b/.github/actions/build-push/action.yml
@@ -1,0 +1,37 @@
+name: Build/push
+description: Build a Docker image and push to ECR
+inputs:
+  aws-access-key-id:
+    description: AWS_ACCESS_KEY_ID
+    required: true
+  aws-secret-access-key:
+    description: AWS_SECRET_ACCESS_KEY
+    required: true
+  aws-region:
+    description: AWS region to use
+    required: true
+    default: us-east-1
+  docker-repo:
+    description: ECR Docker repo to push to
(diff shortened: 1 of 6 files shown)


digitalcora (approved):
Since all feedback was previously addressed in #191, assuming you've tested the deploy process using your fork, this looks fine to me! 🚀

As a side note, if we get this going on a second repo I think it'd be really beneficial in the long run if we could work out how to share some of these generic actions and workflows among all our projects. Since we're building this stuff anew, it would be an ideal time to move away from the "copy and paste the scripts into every repo" approach, and moreso than Semaphore it seems like Actions is designed for this kind of reuse. cc @ianwestcott

paulswartz:
> Since all feedback was previously addressed in #191, assuming you've tested the deploy process using your fork, this looks fine to me! 🚀
> 
> As a side note, if we get this going on a second repo I think it'd be really beneficial in the long run if we could work out how to share some of these generic actions and workflows among all our projects. Since we're building this stuff anew, it would be an ideal time to move away from the "copy and paste the scripts into every repo" approach, and moreso than Semaphore it seems like Actions is designed for this kind of reuse. cc @ianwestcott

Yes, this is absolutely something I want to do. It's much easier to iterate on the actions in one repo/fork, and then pull them out into a shared repo.

PR #196, opened 2021-04-28
use actions from `mbta/actions`

paulswartz:
Also adds deploy notifications:
Screen Shot 2021-04-28 at 3 04 50 PM


paulswartz:
Not sure why I didn't think of it before, but here's an example of a cross-repo PR getting a coverage report: https://github.com/paulswartz/concentrate/pull/6

.github/workflows/dev-green.yml
```diff
@@ -12,16 +12,21 @@ jobs:
...
-      - uses: ./.github/actions/deploy
+      - uses: mbta/actions/deploy-ecs@v1
         with:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           ecs-cluster: concentrate
           ecs-service: concentrate-dev-green-fargate
           docker-tag: ${{ steps.build-push.outputs.docker-tag }}
+      - uses: mbta/actions/notify-slack-deploy@v1
+        if: ${{ !cancelled() }}
```
digitalcora, line 29: If a deploy were cancelled, wouldn't it stop running further steps, and so not reach this point? I'm not sure when this condition would be false.

.github/workflows/dev-green.yml
```diff
@@ -12,16 +12,21 @@ jobs:
     steps:
       - name: Checkout
```
digitalcora, line 13: Unrelated to this PR, but I just noticed this step has a `name`, while the others don't. Necessary? `actions/checkout` seems self-explanatory, if that's what would be shown otherwise.

digitalcora (commented):
A+++++ for extraction of duplicated stuff 😄

.github/workflows/dev-green.yml
paulswartz: From [Job status check functions](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions):
> If your `if` expression does not contain any of the status functions it will automatically result with `success()`. 

So normally yes, but by replacing it with `cancelled()` we'll also run this step if the deploy fails. Example: https://github.com/paulswartz/concentrate/actions/runs/793530029

.github/workflows/dev-green.yml
paulswartz: Sure, I can remove that.

.github/workflows/dev-green.yml
digitalcora: Ohhh, I see. So all steps always run, even if the previous step wasn't successful, but by default have a condition that skips them if previous steps weren't successful. Good to know.

digitalcora (approved):
💯 🍰

PR #198, opened 2021-05-04
ci: use ASDF to install Elixir/Erlang

paulswartz:
This lets us use the single `.tool-versions` configuration for installation, rather than needing to maintain the current version in multiple places. Caching the builds is slow the first time, but quick in future runs.


mkennedm (approved):

mkennedm (approved):
Looks good.

PR #206, opened 2021-06-23
fix: require `stop_sequence` for StopTimeUpdates

digitalcora:
Previously we merged StopTimeUpdates using a "key" of trip ID + stop ID + stop sequence. Since stop ID and sequence are both optional (only one needs to be present), this created the potential for duplicates: two updates that referred to the same stop would not be merged if one of them used a sequence and the other used a stop ID.

Given:

1. it's not universally possible to merge StopTimeUpdates without a stop sequence (it's required if a trip contains a loop)
2. merging updates correctly using "maybe stop ID or sequence or both" would require significant changes to the merging system
3. grouping and the `RemoveUnneededTimes` filter assume the presence of stop sequences
4. all our current input feeds either include stop sequences already or are planned to

...we can simplify things by explicitly requiring a stop sequence, and only merging based on trip ID + stop sequence.

This also includes a refreshed version of an old unmerged feature (#84) that uses static GTFS to fill in stop IDs for updates that only have a stop sequence. Since we are now merging only on stop sequence, it makes sense to not require input feeds to have stop IDs, although this will probably continue to be true for all of our own input feeds.

~~**Note:** If merged, this will effectively disable the `busloc` feed until it is updated to include stop sequences. This might be fine, since attempting to ingest it with the current logic is causing some issues (duplicate updates, departure times at final stops) we might prefer to fix. But, if the current situation is tolerable, we could leave this open until the change is made in `busloc`.~~

### DIFF
--- a/lib/concentrate/filter/gtfs/stop_ids.ex
+++ b/lib/concentrate/filter/gtfs/stop_ids.ex
@@ -1,0 +1,65 @@
+defmodule Concentrate.Filter.GTFS.StopIDs do
+  @moduledoc """
+  Server which knows the stop ID for a given trip ID and stop sequence.
+  """
+  use GenStage
+  alias Concentrate.Filter.GTFS.Helpers
+  require Logger
+  import :binary, only: [copy: 1]
+  @table __MODULE__
+
+  def start_link(opts) do
+    GenStage.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  @spec stop_id(String.t(), non_neg_integer) :: String.t() | :unknown
+  def stop_id(trip_id, stop_sequence) do
+    case :ets.lookup(@table, {trip_id, stop_sequence}) do
+      [{_, value}] -> value
+      [] -> :unknown
+    end
+  rescue
+    ArgumentError -> :unknown
+  end
+
+  @impl GenStage
+  def init(opts) do
+    @table = :ets.new(@table, [:named_table, :public, :set])
+    {:consumer, [], opts}
+  end
+
+  @impl GenStage
+  def handle_events(events, _from, state) do
+    inserts =
+      events
+      |> List.flatten()
+      |> Stream.flat_map(fn
+        {"stop_times.txt", body} ->
+          Helpers.io_stream(body)
+
+        _ ->
+          []
+      end)
+      |> CSV.decode(headers: true, num_workers: System.schedulers())
+      |> Enum.map(&build_insert/1)
+
+    if inserts != [] do
+      true = :ets.delete_all_objects(@table)
+      :ets.insert(@table, inserts)
+
+      Logger.info(fn ->
+        "#{__MODULE__}: updated with #{length(inserts)} records"
+      end)
+    end
+
+    {:noreply, [], state, :hibernate}
+  end
+
+  defp build_insert({:ok, row}) do
+    trip_id = copy(Map.get(row, "trip_id"))
+    stop_id = copy(Map.get(row, "stop_id"))
+    stop_sequence = String.to_integer(Map.get(row, "stop_sequence"))
+
+    {{trip_id, stop_sequence}, stop_id}
+  end
+end
--- a/lib/concentrate/filter/include_stop_id.ex
+++ b/lib/concentrate/filter/include_stop_id.ex
@@ -1,0 +1,34 @@
+defmodule Concentrate.Filter.IncludeStopID do
+  @moduledoc """
+  Adds missing stop IDs to StopTimeUpdates.
+  """
+  @behaviour Concentrate.Filter
+  alias Concentrate.Filter.GTFS.StopIDs
+  alias Concentrate.StopTimeUpdate
+
+  @impl Concentrate.Filter
+  def filter(item, stop_ids \\ StopIDs)
+
+  def filter(%StopTimeUpdate{} = stu, stop_ids) do
+    {:cont,
+     maybe_add_stop_id(
+       stu,
+       StopTimeUpdate.stop_id(stu),
+       StopTimeUpdate.trip_id(stu),
+       StopTimeUpdate.stop_sequence(stu),
+       stop_ids
+     )}
+  end
+
+  def filter(other, _stop_ids), do: {:cont, other}
+
+  defp maybe_add_stop_id(stu, nil, trip_id, stop_sequence, stop_ids)
+       when is_binary(trip_id) and is_integer(stop_sequence) do
+    case stop_ids.stop_id(trip_id, stop_sequence) do
+      :unknown -> stu
+      stop_id -> StopTimeUpdate.update_stop_id(stu, stop_id)
+    end
+  end
+
+  defp maybe_add_stop_id(stu, _, _, _, _), do: stu
+end
--- a/lib/concentrate/filter/null_stop_sequence.ex
+++ b/lib/concentrate/filter/null_stop_sequence.ex
@@ -1,0 +1,19 @@
+defmodule Concentrate.Filter.NullStopSequence do
+  @moduledoc """
+  Filters out StopTimeUpdates with a null `stop_sequence`, since they would not have been merged
+  correctly (and wouldn't work with downstream modules that assume a stop sequence is present).
+  """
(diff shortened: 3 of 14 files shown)


lib/concentrate/filter/null_stop_sequence.ex
```diff
@@ -0,0 +1,17 @@
...
+  Filters out StopTimeUpdates with a null `stop_sequence`, since they would not have been merged
+  correctly (and wouldn't work with downstream modules that assume a stop sequence is present).
+  """
+  @behaviour Concentrate.Filter
+  alias Concentrate.StopTimeUpdate
+  require Logger
+
+  @impl Concentrate.Filter
+  def filter(%StopTimeUpdate{stop_sequence: nil} = stu) do
+    Logger.warn("#{__MODULE__} discarding #{inspect(stu)}")
```
paulswartz, line 12: This seems like it's going to log a ton (every `StopTImeUpdate` in a feed). I'm not sure it's useful in that case?

lib/concentrate/filter/include_stop_id.ex
```diff
@@ -0,0 +1,24 @@
...
+  """
+  @behaviour Concentrate.Filter
+  alias Concentrate.Filter.GTFS.StopIDs
+  alias Concentrate.StopTimeUpdate
+
+  @impl Concentrate.Filter
+  def filter(item, module \\ StopIDs)
+
+  def filter(
+        %StopTimeUpdate{stop_id: nil, trip_id: trip_id, stop_sequence: stop_sequence} = stu,
```
paulswartz, line 13: Concentrate doesn't use pattern matching inside these structs, similarly to how you don't pattern-match on the inside of a `MapSet`. You can match on the struct `def filter(%StopTimeUpdate{} = stu), module)`, but accessing the data should go through the acessor functions.

lib/concentrate/filter/null_stop_sequence.ex
```diff
@@ -0,0 +1,17 @@
...
+  @moduledoc """
+  Filters out StopTimeUpdates with a null `stop_sequence`, since they would not have been merged
+  correctly (and wouldn't work with downstream modules that assume a stop sequence is present).
+  """
+  @behaviour Concentrate.Filter
+  alias Concentrate.StopTimeUpdate
+  require Logger
+
+  @impl Concentrate.Filter
+  def filter(%StopTimeUpdate{stop_sequence: nil} = stu) do
```
paulswartz, line 11: See above comment on pattern-matching the values inside these structs.

lib/concentrate/filter/gtfs/stop_ids.ex
```diff
@@ -0,0 +1,69 @@
...
+  def start_link(opts) do
+    GenStage.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  @spec stop_id(String.t(), non_neg_integer) :: String.t() | :unknown
+  def stop_id(trip_id, stop_sequence) do
+    find_value({trip_id, stop_sequence})
+  end
+
+  defp find_value(key) do
```
paulswartz, line 20: I don't think you need a separate `find_value` here, since `stop_id` is the only user.

paulswartz (commented):
If we're already processing the GTFS files, is it possible to add a filter which guesses at `stop_sequence` for missing values? It wouldn't work 100% of the time (loop routes, as you note), but would allow us to run this without needing to update our source applications to provide `stop_sequence` right away.

digitalcora:
> If we're already processing the GTFS files, is it possible to add a filter which guesses at `stop_sequence` for missing values? It wouldn't work 100% of the time (loop routes, as you note), but would allow us to run this without needing to update our source applications to provide `stop_sequence` right away.

I don't think this would work given the current architecture where filtering happens _after_ merging — a filter that changes the value of a merge key (like stop sequence) would have to run before merging to be useful. I don't know if there's anything really stopping us from adding a new category of "pre-merge filter", but it would be a much heavier lift (from my inexperienced perspective anyway). Alternatively we could add this as a group filter and then "re-merge" the updates within each group, but that seems hacky.

lib/concentrate/filter/include_stop_id.ex
digitalcora: I can certainly change it back — it wasn't clear to me why we were doing that in the original version. What's the reasoning for treating these structs as opaque? When I've seen this used elsewhere, it's like, you're a library used by other applications and don't want _users_ to pattern-match on your structs, so you can change them without breaking compatibility; but here it's all "us" (right?).

lib/concentrate/filter/null_stop_sequence.ex
digitalcora: I think the main point was to let you know somehow _that_ one of your feeds is producing updates with nil stop sequences — if that's valuable then there might be a better way of implementing it, but I think it would involve keeping some kind of state (i.e. we've already issued the warning once for this feed, so don't do it again). We also do already filter out other kinds of "invalid" data silently, so maybe I'll just remove this.

lib/concentrate/filter/include_stop_id.ex
paulswartz: It's "us" inside the `StopTimeUpdate` module, but everything else is "outside". The original idea is that additional filters could be created external to us, while still allowing updates to the internal structure. That hasn't yet happened in practice, but I don't think it should be a big lift to keep new filters using that pattern.

lib/concentrate/filter/null_stop_sequence.ex
paulswartz: It's likely to be obvious in the output that some of the data isn't matching, so I don't think you need this.

lib/concentrate/filter/include_stop_id.ex
```diff
@@ -0,0 +1,33 @@
...
+
+  defp do_filter(stu, nil, trip_id, stop_sequence, stop_ids)
+       when is_binary(trip_id) and is_integer(stop_sequence) do
+    case stop_ids.stop_id(trip_id, stop_sequence) do
+      :unknown -> {:cont, stu}
+      stop_id -> {:cont, StopTimeUpdate.update_stop_id(stu, stop_id)}
+    end
+  end
+
+  defp do_filter(stu, _, _, _, _), do: {:cont, stu}
```
paulswartz, line 32: nit: since all these return `{:cont, stu}` you can pull the `{:cont, }` part up into `filter/2`.

test/concentrate/filter/null_stop_sequence_test.exs
```diff
@@ -0,0 +1,23 @@
+defmodule Concentrate.Filter.NullStopSequenceTest do
+  @moduledoc false
+  use ExUnit.Case, async: true
+  import Concentrate.Filter.NullStopSequence
+  alias Concentrate.StopTimeUpdate
+
+  describe "filter/1" do
+    @tag :capture_log
```
paulswartz, line 8: Do these still need `:capture_log`?

paulswartz (commented):
Maybe I have it wrong in my head, but I think what happens if we deploy this to production is that none of the TripUpdate data from Busloc will be incorporated into Concentrate? Given that is the only way we have to mark updates as SKIPPED (important data we provide to riders) I'd prefer to err on the side of providing slightly incorrect data while we update Busloc to provide stop sequences.

test/concentrate/filter/null_stop_sequence_test.exs
digitalcora: Oops, they should not. Good catch!

digitalcora:
> Maybe I have it wrong in my head, but I think what happens if we deploy this to production is that none of the TripUpdate data from Busloc will be incorporated into Concentrate?

Yep, as I noted in the OP. So we can mark this as blocked for now, I'll find the Busloc task and link it in.

digitalcora:
The `busloc` change that adds stop sequences is now in production, and I deployed this branch again to dev-green to confirm it still merges correctly:

busloc prod | concentrate dev-green
--- | ---
 |

paulswartz (approved):
🍰 now that the Busloc change is out, this should be good to go.

PR #220, opened 2021-11-18
docs: README change of file path for symbolic linking pre-commit under hooks

uchukwu:
README change of file path for symbolic linking of pre-commit under hooks

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ asdf install
 mix deps.get
 
 # add pre-commit hook to verify formatting/tests/types
-ln -s ../../hooks/pre-commit .git/hooks/pre-commit
+ln -s ./hooks/pre-commit .git/hooks/pre-commit
 
 # make sure everything passes! (slowest to fastest)
 mix format --check-formatted
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ asdf install
 mix deps.get
 
 # add pre-commit hook to verify formatting/tests/types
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ asdf install
 mix deps.get
 
 # add pre-commit hook to verify formatting/tests/types
-ln -s ../../hooks/pre-commit .git/hooks/pre-commit
+ln -s ./hooks/pre-commit .git/hooks/pre-commit
```
bklebe, line 26:
```suggestion
ln -s hooks/pre-commit .git/hooks/pre-commit
```

bklebe (approved):

paulswartz (changes requested):
@bklebe @uchukwu does doing the link this way run the tests for you? It doesn't on my machine.

bklebe:
> @bklebe @uchukwu does doing the link this way run the tests for you? It doesn't on my machine.

I should have tested this, it doesn't seem to work. I approved it because it looked right from where the hooks file is, but I assume it's because the symbolic link needs to be relative to the .git folder.

bklebe:
As an alternative you could use the absolute path to make this instruction less confusing, i.e. `ln -s "$(pwd)/hooks/pre-commit .git/hooks/pre-commit`. Or it looks like you can set `git config core.hooksPath hooks/` from the root of the working tree as well.

bklebe (changes requested):
requires changes, per @paulswartz

PR #237, opened 2022-05-12
fix: align ECS service name with Terraform

bklebe:
When https://github.com/mbta/devops/pull/381 is applied, the dev green service will be renamed to concentrate-dev-green. This PR aligns the deployment workflow with that change.


losvedir (approved):
Looks good. Just ran the deploy from this branch and it went through and looks good. 👍

PR #242, opened 2022-06-10
deps: update to latest release of Erlang 23

digitalcora:
The project's previous version of Erlang no longer builds on current versions of macOS, which is a blocker for continued work on it.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-elixir 1.10.3-otp-23
-erlang 23.0.2
+elixir 1.10.4-otp-23
+erlang 23.3.4.14
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM hexpm/elixir:1.10.3-erlang-23.0.2-alpine-3.11.6 AS builder
+FROM hexpm/elixir:1.10.4-erlang-23.3.4.14-alpine-3.16.0 AS builder
 
 WORKDIR /root
 
@@ -22,7 +22,7 @@ ADD src /root/src
 RUN mix do compile, release
 
 # Second stage: copies the files from the builder stage
-FROM alpine:3.11.6
+FROM alpine:3.16.0
 
 RUN apk add --update libssl1.1 ncurses-libs bash dumb-init \
     && rm -rf /var/cache/apk


.tool-versions
```diff
@@ -1,2 +1,2 @@
-elixir 1.10.3-otp-23
-erlang 23.0.2
+elixir 1.10.4-otp-23
+erlang 23.2.3
```
paulswartz, line 2: question: why not upgrade to the latest point release, 23.3.4.14?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM hexpm/elixir:1.10.3-erlang-23.0.2-alpine-3.11.6 AS builder
+FROM hexpm/elixir:1.10.4-erlang-23.2.3-alpine-3.12.1 AS builder
```
paulswartz, line 1: question: if you're upgrading Alpine, why not upgrade to the latest version (3.16.0)?

digitalcora:
@paulswartz re: questions, the answer is just that I wanted to be as conservative as possible to reduce the risk of getting hung up on any issues unrelated to the real task I'm looking to start. If you're confident those further upgrades would be just as "safe" as these, then sure.

paulswartz:
@digitalcora I'm pretty confident that those changes would be safe. 

If you're trying to be conservative, I'd upgrade to Erlang 23.0.4 instead, and to Alpine 3.13 (3.12 is [EOLed https://endoflife.date/alpine] anyways): https://hub.docker.com/layers/elixir/hexpm/elixir/1.10.4-erlang-23.0.4-alpine-3.13.1/images/sha256-fccb96a73fe844a95e1c8b8cc37a2f5295af74373a78449fb58030d7e4ecf576?context=explore

digitalcora:
I did try 23.0.4, but unfortunately it also doesn't build on current macOS. From a bit of searching, 23.2.1 might be the first release that fixes the issue, but I just went with 23.2.3 because that was what RTR was already using.

Anyway, I'll go ahead and try those other versions.

digitalcora:
This seems to be running fine on dev-green. I used 23.3.4.9 instead of 23.3.4.14, since it was the latest version available for Elixir 1.10.4 in the Hexpm Docker images — as per [their readme](https://hub.docker.com/r/hexpm/elixir) they don't build new images for sub-patch releases.

paulswartz:
@digitalcora is https://hub.docker.com/layers/elixir/hexpm/elixir/1.10.4-erlang-23.3.4.14-alpine-3.16.0/images/sha256-d42190f2903c92f4055f3d6057b8acd76822d1f7e7599fc4cca2ba463fa75c59?context=explore not the right image?

digitalcora:
Ah, thanks, that didn't show up in my tag search — at first the results appeared to be sorted with the newest versions at the top, so I assumed .9 was the last sub-patch they had. Scrolling down a bit, I see it's more complicated. Time to build another version, whee~

digitalcora:
@paulswartz Once more with feeling!

paulswartz (approved):
:cake: assuming all looks good on dev-green.

PR #243, opened 2022-06-22
feat: filter to fill in stop times from schedules

digitalcora:
**Asana:** [🧠 Provide our own predictions for some CR trains](https://app.asana.com/0/584764604969369/1202190609541062)

Best reviewed per-commit, as there's a noisy refactor beforehand.

⚠️ Testing this in a deployed environment is currently blocked on an issue where the CRB `dev` environment has not actually updated its feed since 5/25 ([Asana](https://app.asana.com/0/881264583703207/1202482229913755)). Since this means all of its updates are very outdated, they get filtered out, and we can't see the effect of the new filter.

**Configurability** is in a separate WIP commit because I was not sure whether this was the best way to go about it. Hard-coding MBTA-specific status values into the filter didn't seem ideal, but there are some open questions in my mind:

* Is it fine to hard-code `status` as the field that triggers this behavior, or do we want the key to be configurable as well as the values?
* Is the slight metaprogramming/testing awkwardness worth it to allow the values to be known at compile time? Perhaps it would also be fine to use a runtime MapSet membership test.


lib/concentrate/group_filter/scheduled_stop_times.ex
```diff
@@ -0,0 +1,77 @@
+defmodule Concentrate.GroupFilter.ScheduledStopTimes do
+  @moduledoc """
+  Uses the static GTFS schedule to fill in missing arrival/departure times on stop time updates
+  that have specific `status` values.
+
+  The desired status values must be set in app configuration. If no status values are configured,
```
paulswartz, line 6: A run-time configuration of the filters will not affect this filter (as currently written):
```suggestion
  The desired status values must be set in app configuration at compile time. If no status values are configured,
```

test/concentrate/group_filter/scheduled_stop_times_test.exs
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  config_path = [:group_filters, ScheduledStopTimes, :on_time_statuses]
+  @on_time_statuses Application.compile_env(:concentrate, config_path, [])
+
+  defp filter(trip_group), do: ScheduledStopTimes.filter(trip_group, FakeStopTimes)
+  defp on_time_status, do: Enum.random(@on_time_statuses)
+
+  if @on_time_statuses == [] do
+    @moduletag skip: "not testable when no on-time status values are configured"
```
paulswartz, line 21: question: can't you set this configuration in `config/test.exs`?

paulswartz (commented):
> * Is it fine to hard-code `status` as the field that triggers this behavior, or do we want the key to be configurable as well as the values?

Given that adding additional fields would require updating the GTFS-RT parsers, I don't think it's worth having that be programmable.
 
> * Is the slight metaprogramming/testing awkwardness worth it to allow the values to be known at compile time? Perhaps it would also be fine to use a runtime MapSet membership test.

I think a runtime test would be better, so that these can be configured at runtime with the JSON file. Otherwise, we'd need to do a code update if the status values change, instead of only a configuration change.

digitalcora:
> Given that adding additional fields would require updating the GTFS-RT parsers, I don't think it's worth having that be programmable.

I meant only fields that are already part of StopTimeUpdate; for example someone could have this behavior trigger on specific values of `uncertainty` or `schedule_relationship`. (I don't know why they would do that, but they could)

> I think a runtime test would be better, so that these can be configured at runtime with the JSON file.

Any suggestions for what this should look like, structurally? Currently we don't seem to have anything like a "filter configuration" in the runtime config. Should there be a new top-level key?

test/concentrate/group_filter/scheduled_stop_times_test.exs
digitalcora: Ah, yes, definitely. Just overthunk it.

lib/concentrate/gtfs/stop_times.ex
```diff
@@ -0,0 +1,134 @@
...
+
+  defp build_inserts(_, _), do: []
+
+  @twelve_hours 12 * 60 * 60
+
+  # Convert a GTFS "time" to an offset in seconds from noon, which as per the spec is what GTFS
+  # times actually mean (rather, they are an "offset from 12 hours before noon"; we pre-subtract
+  # the 12-hour offset as well).
+  defp time_to_offset(
+         <<hour::binary-size(2), ":", minute::binary-size(2), ":", second::binary-size(2)>>
```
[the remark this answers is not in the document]
digitalcora: The spec does say "HH:MM:SS format", but doesn't specifically say that leading zeros are required. `gtfs_kit` validators accepts H:MM:SS, as well. So maybe we should support it.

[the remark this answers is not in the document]
paulswartz: There's a [GTFS PR](https://github.com/google/transit/pull/328/files) which updates the language to include "H:MM:SS", so even that might change sometime soon.

config/config.exs
```diff
@@ -30,6 +32,12 @@ config :concentrate,
     Concentrate.Filter.IncludeStopID
   ],
   group_filters: [
+    {
+      Concentrate.GroupFilter.ScheduledStopTimes,
+      # https://github.com/mbta/commuter_rail_boarding/blob/79a493f/config/config.exs#L34-L63
+      on_time_statuses: ["All aboard", "Now boarding", "On time", "On Time"]
+    },
+    Concentrate.GroupFilter.ScheduledStopTimes,
```
paulswartz, line 40: question: does this need to be in here twice?

lib/concentrate/group_filter/scheduled_stop_times.ex
```diff
@@ -0,0 +1,77 @@
...
+  enabling this filter has no effect. Example:
```
paulswartz, line 7:
```suggestion
  enabling this filter has no effect. 
  
  Example:
```

Previous version reads like this is an example of when it has no effect.

lib/concentrate/group_filter/scheduled_stop_times.ex
```diff
@@ -0,0 +1,77 @@
+defmodule Concentrate.GroupFilter.ScheduledStopTimes do
+  @moduledoc """
+  Uses the static GTFS schedule to fill in missing arrival/departure times on stop time updates
+  that have specific `status` values.
```
paulswartz, line 4: question: should this note that it needs to come before `RemoveUnneededTimes` in the filter list, or not include arrival/departure times in cases when we'd drop them?

paulswartz (commented):
> > Given that adding additional fields would require updating the GTFS-RT parsers, I don't think it's worth having that be programmable.
> 
> I meant only fields that are already part of StopTimeUpdate; for example someone could have this behavior trigger on specific values of `uncertainty` or `schedule_relationship`. (I don't know why they would do that, but they could)

Hmm. `uncertainty` might be interesting. I know Swiftly uses specific values in their output: https://swiftly-inc.stoplight.io/docs/realtime-standalone/b3A6Mjg0MzYyMTk-gtfs-rt-trip-updates

However, it's probably YAGNI, so I wouldn't worry about it.
 
> > I think a runtime test would be better, so that these can be configured at runtime with the JSON file.
> 
> Any suggestions for what this should look like, structurally? Currently we don't seem to have anything like a "filter configuration" in the runtime config. Should there be a new top-level key?

On second thought: I think making the filter configuration run-time configurable in general would be nice, but I don't think it needs to be part of this PR. Can you create a new Asana task for that (if one doesn't already exist)?

New question: were you able to configure Concentrate locally and see the correct behavior in its output?

lib/concentrate/group_filter/scheduled_stop_times.ex
digitalcora: Hm, good point.

lib/concentrate/group_filter/scheduled_stop_times.ex
digitalcora: I don't think it _needs_ to, strictly speaking. It does need to if you want `RemoveUnneededTimes` to remove times it generates, but I think that's a logical consequence of it appearing later in the filter list. If it's generally understood that filters are run in the order they're configured, I think we don't have to specifically call it out here. (Maybe we do want to call that out somewhere in the general documentation?)

digitalcora:
> New question: were you able to configure Concentrate locally and see the correct behavior in its output?

I've just done this, and the good news is that the filter works as intended. The bad news is that `StopTimes` takes 4 minutes (at least on my machine) to populate the ETS table, and during most of that time the filter is inoperative since `StopTimes` deletes all records before starting to insert the replacements. The approach here was mostly copied from `PickupDropOff`, which I notice also has this problem, taking ~1.5 minutes to populate even when `StopTimes` is disabled. However, looking at `StopIDs`, I can at least improve the approach so the inserts are done all at once (and maybe apply that same approach to `PickupDropOff` as a refactor).

This still leaves a few-second gap between the delete and the insert, so unfortunately an incorrect feed might be produced during that time. I wonder if it would be worth a refactor to either use an "upsert" approach, or use Mnesia as `api` does. Maybe not as part of this task, since the problem is not new to this task.

digitalcora:
> It looks like `PickupDropOff` parses the CSV in parallel: does using that speed up the ingest for Schedule?

If you mean for the new `StopTimes` here, it does already do that (see L102).

digitalcora:
> I don't suppose you tried using eflame or similar to see where the time is going?

I haven't yet done so, since I thought the memory usage would be more of an issue than the time usage, though I guess the time is also an issue for the initial warm-up of the app (given it's single-instance). I think what I'll do is try a quick experiment with combining all the stop-times servers, and see what that does for both metrics. If the warm-up time is no _worse_ than it is currently, I'd be comfortable declaring the issue out-of-scope for this task.

digitalcora:
Good news on the experiment: Memory usage is ~2.0GB at max, ~880MB steady state, and the update takes ~1.5 minutes, the same time as `PickupDropOff` previously took alone. I'll clean up this approach and add it as a refactor commit, then we can think about further optimizations later since the performance is not made any worse by this PR.

digitalcora:
Refactor done! There are some incidental changes here that should not have any effect on the behavior of the app for our purposes (I also documented them in the commit message):

* `RemoveUnneededTimes` no longer works with stop time updates that are missing a `stop_sequence`. This should not have happened anyway since ac72186, which formalized the requirement that all stop time updates should have a stop sequence for merging to work correctly.

* `RemoveUnneededTimes` and `IncludeStopID` no longer work with GTFS feeds that have _only_ a `stop_times.txt`. Like `ScheduledStopTimes`, they now require the full chain of `stop_times.txt`, `trips.txt`, `routes.txt`, and `agencies.txt` to be present and valid, though the files are unused in these filters. Both the old and new behaviors are accidents of the implementation details, and the new behaviour could be changed if we wanted to explicitly support invalid "partial" GTFS feeds, but as far as I know we have no need for this.

losvedir:
I pushed up a small commit that resolved an issue I think from a last minute refactor where `StopTimes.get` no longer exists (now it's `StopTimes.arrival_departure`). After that, it appears to work!

Here's dotcom-dev-green with all its boarding glory:

Screen Shot 2022-07-20 at 3 53 23 PM

Compared to prod which is all "Scheduled":

Screen Shot 2022-07-20 at 3 53 36 PM

The metrics in ECS seem fine, so I think this branch is a go!

losvedir (approved):
🎉

PR #248, opened 2022-07-21
feat: filter schedule based vehicles

bfauble:
Asana: [🧠 Address Swiftly "schedule-based vehicle" IDs](https://app.asana.com/0/505721188639414/1201938900646058/f)


losvedir (changes requested):
I love that you were able to dive in and 1) make a new filter and 2) make it dynamic via `config.exs`. So, I'd say the core of the approach here is solid.

However, and this is my bad for not being clearer in the ticket, the desired functionality is actually updating the `TripUpdates` (i.e.: "predictions") output, not the `VehiclePositions`.

If you look at [concentrate's output](https://mbta-gtfs-s3.s3.amazonaws.com/concentrate/TripUpdates_enhanced.json) for `TripUpdates`, you'll see that each TripUpdate entity has a `vehicle` field, which is supposed to say which vehicle is serving that trip. However, Swiftly - the vendor that makes predictions for us for Bus and Commuter Rail - sometimes makes predictions before it knows what vehicle is going to do the trip, based on the schedule.

For instance the schedule might say that a bus will be leaving Harvard in 10 minutes, but no bus or driver has actually signed in for that trip yet. The options here are to just _not_ make predictions, even though _usually_ the bus and driver will show up and the trip will happen, or to make predictions based on the schedule. Swiftly does the latter, and when it does that, since it doesn't know the ID of the actual bus that will be doing the trip, it just makes up an ID! However, that vehicle ID is invalid, since it's supposed to refer to a real vehicle, and that vehicle should be in our `VehiclePositions` feed.

So what we actually want is a filter on `TripDescriptor` that drops the Vehicle ID if it matches the pattern.

Separately, and a smaller change, I'd suggest the name be something like `FilterTripUpdateVehicles` or something that is 1) clearer that it's the TripUpdates that are being affected, and 2) more generic in the reasoning - for us, it's because we have `schedBasedVehicles` we want removed, but other hypothetical users of concentrate (or us in the future) might have other invalid vehicle IDs they want removed.

bfauble:
> So what we actually want is a filter on `TripDescriptor` that drops the Vehicle ID if it matches the pattern.

By this you mean null out the field vs remove the entry entirely?

losvedir:
> By this you mean null out the field vs remove the entry entirely?

Yes, Trip Updates are not required to have vehicle information, and we still want to keep Swiftly's schedule based predictions. We just don't want to publish a fake Vehicle ID.

bklebe (approved):
Looks good to me, pending Gabe's confirmation that this is what he intended.

lib/concentrate/filter/filter_trip_update_vehicles.ex
```diff
@@ -0,0 +1,35 @@
...
+  @behaviour Concentrate.Filter
+
+  config_path = [:filters, __MODULE__, :suffix_matches]
+  @suffix_matches Application.compile_env(:concentrate, config_path, [])
+
+  @impl Concentrate.Filter
+  def filter(td, suffix_matches \\ @suffix_matches)
+  def filter(td, []), do: {:cont, td}
+
+  def filter(%TripDescriptor{trip_id: trip_id} = td, suffix_matches) do
```
losvedir, line 26: Should be `vehicle_id`, no?

bfauble: yep you're right, will update

test/concentrate/filter/filter_trip_update_vehicles_test.exs
```diff
@@ -0,0 +1,46 @@
...
+      assert {:cont, %TripDescriptor{trip_id: "trip1", vehicle_id: nil}} =
+               filter(td1, ["suffix_1", "suffix_2", "suffix_3"])
+
+      assert {:cont, %TripDescriptor{trip_id: "trip2", vehicle_id: nil}} =
+               filter(td2, ["suffix_1", "suffix_2", "suffix_3"])
+
+      assert {:cont, %TripDescriptor{trip_id: "trip3", vehicle_id: nil}} =
+               filter(td3, ["suffix_1", "suffix_2", "suffix_3"])
+
+      assert {:cont, ^td4} = filter(td4, ["suffix_1", "suffix_2", "suffix_2"])
```
losvedir, line 39: I think this last element of the list should be `"suffix_3"`.

losvedir:
Looks good to me! Can you make sure your branch is rebased on the latest `master` and then deploy to dev-green to run for a little while, so we can make sure the logs look good?

lib/concentrate/filter/filter_trip_update_vehicles.ex
```diff
@@ -0,0 +1,37 @@
...
+  config_path = [:filters, __MODULE__, :suffix_matches]
+  @suffix_matches Application.compile_env(:concentrate, config_path, [])
+
+  @impl Concentrate.Filter
+  def filter(td, suffix_matches \\ @suffix_matches)
+  def filter(td, []), do: {:cont, td}
+
+  def filter(%TripDescriptor{vehicle_id: nil} = td, _), do: {:cont, td}
+
+  def filter(%TripDescriptor{vehicle_id: vehicle_id} = td, suffix_matches) do
```
paulswartz, line 28: suggestion: generally we don't pattern match or directly update the internals of the GTFS-RT structs: https://github.com/mbta/concentrate/blob/master/lib/concentrate/filter/include_stop_id.ex

You could do this in a similar way, however, by pulling the vehicle ID out into a variable and passing it to a new function:
```elixir
def filter(%TripDescriptor{} = td) do
  vehicle_id = TripDescriptor.vehicle_id(td)
  td = possibly_strip_vehicle_id(vehicle_id, td, suffix_matches)
  {:cont, td}
end
def filter(other, _) do
  {:cont, other}
end

defp possibly_strip_vehicle_id(nil, td, _), do: td
defp possibly_strip_vehicle_id(vehicle_id, td, suffix_matches) do
  if String.ends_with?(vehicle_id, suffix_matches) do
    TripDescriptor.update_vehicle_id(td, nil)
  else
    td
  end
end
```

paulswartz (approved):
🍰

losvedir (approved):
Finally had a chance to play around with the actual output here.

I confirmed that prod had a number of `schedBasedVehicle` vehicle IDs, and that dev-green had none. I also looked at the TripUpdate entity in dev-green corresponding to one that had that vehicle ID and confirmed that it otherwise looked good, and just lacked the `"vehicle"` block, as expected.

losvedir:
I think this is probably most conceptually just a single commit. Can you squash it when you merge?

PR #263, opened 2022-12-28
Erlang 24 upgrade

nlwstein:
#### Summary of changes

Upgrades project to Erlang 24 and Elixir 1.11, so that it can be built on Ubuntu 22.04. This is an alternative approach to [using Docker to run the tests](https://github.com/mbta/concentrate/pull/262). 

Main concern is [this fix](https://github.com/mbta/concentrate/pull/263/files#diff-2c04e9646ab4e9b75b4af509112dcca2509622bbaaeaf5be645d03b583edcb97R16) - is it a problem that encoding and decoding the list results in a slight reordering? I would guess not, but it's a guess 😄 

**Asana Ticket:** [Concentrate CI runs on Ubuntu 22.04](https://app.asana.com/0/584764604969369/1203566390446235/f)

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -1,6 +1,8 @@
 name: Deploy to Dev
 
 on:
+  schedule:
+    - cron: '0 5 * * *'
   workflow_dispatch:
   push:
     branches: [master]
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -1,8 +1,6 @@
 name: Deploy to Dev
 
 on:
-  schedule:
-    - cron: '0 5 * * *'
   workflow_dispatch:
   push:
     branches: [master]
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -44,7 +44,7 @@ jobs:
         mix local.hex --force
         mix deps.get
     - name: Compile (warnings as errors)
-      run: mix compile --force --warnings-as-errors
+      run: mix compile --force
     - name: Check formatting
       run: mix format --check-formatted
     - name: Credo
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -1,6 +1,10 @@
 name: Elixir CI
 
-on: [push, pull_request]
+on: 
+  push:
+  pull_request:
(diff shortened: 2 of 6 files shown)


.github/workflows/dev.yml
```diff
@@ -1,6 +1,8 @@
 name: Deploy to Dev
 
 on:
+  schedule:
+    - cron: '0 5 * * *'
```
bklebe, line 5: Would you mind moving these cron changes to a separate PR?

bklebe (changes requested):

.github/workflows/dev.yml
nlwstein: @bklebe All set! PR - https://github.com/mbta/concentrate/pull/264

.github/workflows/elixir.yml
```diff
@@ -44,7 +46,7 @@ jobs:
         mix local.hex --force
         mix deps.get
     - name: Compile (warnings as errors)
-      run: mix compile --force --warnings-as-errors
+      run: mix compile --force
```
bklebe, line 49: Sorry I didn't mention this before - could you make an inbound ticket to fix the compiler warnings here and turn `--warnings-as-errors` back on?

bklebe (commented):
looks good Nick — would you mind cleaning up your commit history using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)? If you prefer you could also just squash them all into a `chore: update to Erlang 24` commit before you merge in GitHub too.

.github/workflows/elixir.yml
nlwstein: > Sorry I didn't mention this before - could you make an inbound ticket to fix the compiler warnings here and turn `--warnings-as-errors` back on?

@bklebe I can do this, but it will cause the workflow to get marked as failed.

bklebe: No, I mean you can leave this the way it is, just make a ticket to fix the warnings.

nlwstein: Sorry - may have misunderstood what you said - do you want the ticket to also be about turning this feature back on? I have created one for that: https://app.asana.com/0/881264583703207/1203676093346059/f

PR #277, opened 2023-04-06
Temporarily suppress terminal/reverse trip predictions for Red Line and Blue Line in Concentrate

bklebe:
#### Summary of changes

**Asana Ticket:** [[extra] 🧠 Temporarily suppress terminal/reverse trip predictions for Red Line and Blue Line in Concentrate](https://app.asana.com/0/584764604969369/1204348457726555/f)

Adds a new GroupFilter that removes uncertain StopTimeUpdates from a group that match a route_id and uncertainty value specified as a map in `config.exs`. Also updates `config.exs` to suppress Red and Blue line terminal and reverse predictions while we wait for schedules to recover.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -37,6 +37,10 @@ config :concentrate,
   ],
   group_filters: [
     {
+      Concentrate.GroupFilter.RemoveUncertainStopTimeUpdates,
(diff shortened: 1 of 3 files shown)


paulswartz (approved):
:cake: thanks!

PR #281, opened 2023-06-05
Don't mix predictions

Whoops:
#### Summary of changes

**Asana Ticket:** 
[🧠 🐛 Don't mix predictions: Multiple sources to Swiftly, same trip](https://app.asana.com/0/584764604969369/1204277338626227/f)
[🧠 🐛 Don't mix predictions: Multiple buses, same trip](https://app.asana.com/0/584764604969369/1204277338626221/f)

This ensures that we are always using times from the same prediction when merging StopTimeUpdates and filters trips that suggest we'll arrive at a later stop before departing a previous one.

### DIFF
--- a/lib/concentrate/group_filter/time_travel.ex
+++ b/lib/concentrate/group_filter/time_travel.ex
@@ -28,6 +28,7 @@ defmodule Concentrate.GroupFilter.TimeTravel do
       Logger.warning("""
       Trip ID: #{stop_time_update.trip_id} predicts arriving at stop #{stop_time_update.stop_sequence} at #{time}
       before departing stop #{prev.stop_sequence} at #{prev_time}. Dropping prior predictions.
+      trip_id=#{stop_time_update.trip_id} stop_sequence=#{stop_time_update.stop_sequence}
       """)
 
       [stop_time_update]
--- a/lib/concentrate/group_filter/time_travel.ex
+++ b/lib/concentrate/group_filter/time_travel.ex
@@ -7,32 +7,32 @@ defmodule Concentrate.GroupFilter.TimeTravel do
   @behaviour Concentrate.GroupFilter
   @impl Concentrate.GroupFilter
 
-  def filter({td, vps, stus}) do
-    stus =
-      stus
-      |> Enum.reduce([], &filter_stu/2)
+  def filter({td, vps, stop_time_updates}) do
+    stop_time_updates =
+      stop_time_updates
+      |> Enum.reduce([], &filter_stop_time_update/2)
       |> Enum.reverse()
 
-    {td, vps, stus}
+    {td, vps, stop_time_updates}
   end
 
-  defp filter_stu(stu, []) do
-    [stu]
+  defp filter_stop_time_update(stop_time_update, []) do
+    [stop_time_update]
   end
 
-  defp filter_stu(stu, [prev | _] = stus) do
+  defp filter_stop_time_update(stop_time_update, [prev | _] = stop_time_updates) do
     prev_time = prev.departure_time || prev.arrival_time
-    time = stu.arrival_time || stu.departure_time
+    time = stop_time_update.arrival_time || stop_time_update.departure_time
 
     if time < prev_time do
       Logger.warning("""
-      Trip ID: #{stu.trip_id} predicts arriving at stop #{stu.stop_sequence} at #{time}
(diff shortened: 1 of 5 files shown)


lib/concentrate/group_filter/time_travel.ex
```diff
@@ -0,0 +1,38 @@
...
+      |> Enum.reverse()
+
+    {td, vps, stus}
+  end
+
+  defp filter_stu(stu, []) do
+    [stu]
+  end
+
+  defp filter_stu(stu, [prev | _] = stus) do
```
nlwstein, line 23: This is really minor but it might be more readable to use `stop_time_update` in place of `stu` (etc) in this file. Not blocking, just an opinion 😄

test/concentrate/group_filter/time_travel_test.exs
```diff
@@ -0,0 +1,80 @@
...
+      ])
+
+    {td, vp, stus} = expected_trip
+    actual_trip = TimeTravel.filter(expected_trip)
+    assert {^td, ^vp, stus} = actual_trip
+
+    assert trip_stops(actual_trip) == trip_stops(expected_trip)
+  end
+
+  test "drops predecing stops when prediction involves going back in time" do
```
nlwstein, line 50: nit: did you mean `preceding`?

lib/concentrate/group_filter/time_travel.ex
```diff
@@ -0,0 +1,38 @@
...
+    prev_time = prev.departure_time || prev.arrival_time
+    time = stu.arrival_time || stu.departure_time
+
+    if time < prev_time do
+      Logger.warning("""
```
nlwstein, line 28: Not blocking - If you use key-value pairs like `trip_id=SOME_VALUE some_other_key=some_other_value ...` in your logs, they will be parsed by Splunk as data. Sometimes it's helpful to append this to the end of your human readable string if you want to filter these error messages quickly.

nlwstein (changes requested):
The only thing that should prob be fixed before merging is the spelling q 😄

lib/concentrate/stop_time_update.ex
```diff
@@ -38,10 +38,12 @@ defmodule Concentrate.StopTimeUpdate do
     def key(%{trip_id: trip_id, stop_sequence: stop_sequence}), do: {trip_id, stop_sequence}
 
     def merge(first, second) do
+      [first, second] = Enum.sort_by([first, second], &Concentrate.StopTimeUpdate.time/1)
```
paulswartz, line 41: question: this is adding quite a few calls to a frequently called codepath: what's the performance like?

lib/concentrate/group_filter/time_travel.ex
Whoops: Agreed. `stu` was the convention used elsewhere, but it also confused me, so I have changed it to be more verbose.

test/concentrate/group_filter/time_travel_test.exs
Whoops: I did.

lib/concentrate/group_filter/time_travel.ex
Whoops: That's useful, I've added `trip_id` and `stop_sequence`

lib/concentrate/stop_time_update.ex
Whoops: I was hoping to put real world numbers but after profiling Concentrate for 5 minutes I was surprised to find it wasn't called at all! So putting numbers to it a different way...

```elixir
iex(1)> base_stu = %Concentrate.StopTimeUpdate{stop_id: 1} 
...
iex(2)> stus = Enum.map(1..1000000, fn _ -> %{base_stu | arrival_time: :random.uniform(100), departure_time: :random.uniform(100)} end)
...
iex(3): timer.tc(fn -> Enum.reduce(stus, &Concentrate.Mergeable.merge/2) end)

# 3 runs
# Enum.sort_by version
#1: 225,835
#2: 205,168
#3: 259,574

# If statement version
#1: 139,680
#2: 140,282
#3: 140,623
```

After writing it I think the `if` is about as clear as the sort, so I've updated to use that.

nlwstein (approved):

nlwstein:
This looks good to me, once the thread w/ @paulswartz is resolved 😄

paulswartz (approved):

PR #282, opened 2023-06-09
feat: Keolis realtime gtfs

Whoops:
#### Summary of changes

**Asana Ticket:** [🚆Keolis Feed Transition](https://app.asana.com/0/584764604969369/1204680576534832/f)

Does 3 main things:
1) Adds the Keolis GTFS enhanced feeds to the local configuration
2) Makes the handling of on time statuses no longer case sensitive
3) Overrides the boarding statuses provided by Keolis to match those in Commuter Rail Boarding

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -16,6 +16,12 @@ config :concentrate,
     gtfs_realtime: [
       vehicle_positions: "https://cdn.mbta.com/realtime/VehiclePositions.pb",
       trip_updates: "https://cdn.mbta.com/realtime/TripUpdates.pb"
+    ],
+    gtfs_realtime_enhanced: [
+      keolis_trip_updates:
+        "https://mbta-gtfs-s3.s3.amazonaws.com/keolis/dev/TripUpdates_enhanced.json",
+      keolis_vehicle_positions:
+        "https://mbta-gtfs-s3.s3.amazonaws.com/keolis/dev/VehiclePositions_enhanced.json"
     ]
   ],
   alerts: [
--- a/config/config.exs
+++ b/config/config.exs
@@ -12,6 +12,28 @@ config :ex_aws, json_codec: Jason
 config :ssl, protocol_version: :"tlsv1.2"
 
 config :concentrate,
+  boarding_status_override: %{
+    "ARRIVED" => "Arrived",
+    "CANCELLED" => "Cancelled",
+    "DELAYED" => "Delayed",
+    "DEPARTED" => "Departed",
+    "BOARDING COMPLETE" => "Boarding complete",
+    "ON TIME" => "On time",
+    "PRIORITY" => "Info to follow",
+    "BUS SUBSTITUTE" => "Bus substitution",
+    "SEE AGENT" => "See agent",
+    "ORANGE LINE" => "Not stopping here",
+    "GREEN LINE" => "Not stopping here",
+    "RED LINE" => "Not stopping here",
+    "BLUE LINE" => "Not stopping here",
+    "SILVER LINE" => "Not stopping here",
+    "SUBWAY" => "Not stopping here",
+    "NOW BOARDING" => "Now boarding",
+    "ALL ABOARD" => "All aboard",
+    "ARRIVING" => "Arriving",
+    "LATE" => "Late",
+    "HOLD" => "Info to follow"
+  },
   sources: [
     gtfs_realtime: [
       vehicle_positions: "https://cdn.mbta.com/realtime/VehiclePositions.pb",
--- a/config/config.exs
+++ b/config/config.exs
@@ -38,12 +38,6 @@ config :concentrate,
     gtfs_realtime: [
       vehicle_positions: "https://cdn.mbta.com/realtime/VehiclePositions.pb",
       trip_updates: "https://cdn.mbta.com/realtime/TripUpdates.pb"
-    ],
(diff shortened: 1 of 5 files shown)


config/config.exs
```diff
@@ -12,10 +12,38 @@ config :ex_aws, json_codec: Jason
 config :ssl, protocol_version: :"tlsv1.2"
 
 config :concentrate,
+  boarding_status_override: %{
+    "ARRIVED" => "Arrived",
+    "CANCELLED" => "Cancelled",
+    "DELAYED" => "Delayed",
+    "DEPARTED" => "Departed",
+    "BOARDING COMPLETE" => "Boarding complete",
```
Whoops, line 20: These are built off of: https://github.com/mbta/commuter_rail_boarding/blob/main/config/config.exs#L34

"BOARDING COMPLETE" is the only status that does not appear to have an equivalent status in Commuter Rail Boarding.

config/config.exs
```diff
@@ -12,10 +12,38 @@ config :ex_aws, json_codec: Jason
...
+    "ARRIVING" => "Arriving",
+    "LATE" => "Late",
+    "HOLD" => "Info to follow"
+  },
   sources: [
     gtfs_realtime: [
       vehicle_positions: "https://cdn.mbta.com/realtime/VehiclePositions.pb",
       trip_updates: "https://cdn.mbta.com/realtime/TripUpdates.pb"
+    ],
+    gtfs_realtime_enhanced: [
```
paulswartz, line 42: suggestion: this configuration is the default, and other potential users of Concentrate don't have access to these feeds. Instead, they'll go into the configuration in Terraform.

lib/concentrate/group_filter/scheduled_stop_times.ex
```diff
@@ -58,18 +62,22 @@ defmodule Concentrate.GroupFilter.ScheduledStopTimes do
     end
 
     defp fill_in_arrival_departure(stop_time_update, trip_date, nil, nil, status, gtfs_stop_times)
-         when status in @on_time_statuses do
-      trip_id = StopTimeUpdate.trip_id(stop_time_update)
-      stop_sequence = StopTimeUpdate.stop_sequence(stop_time_update)
+         when is_binary(status) do
+      if String.downcase(status) in @on_time_statuses do
```
paulswartz, line 66: question: what's the performance of this change? It used to be handled with pattern matching, but it doesn't look like that's possible with the implementation here.

lib/concentrate/parser/gtfs_realtime_enhanced.ex
```diff
@@ -122,6 +127,14 @@ defmodule Concentrate.Parser.GTFSRealtimeEnhanced do
     end
   end
 
+  defp decode_boarding_status(nil) do
+    nil
+  end
+
+  defp decode_boarding_status(status) do
+    Map.get(@boarding_status_override, status, status)
```
paulswartz, line 135: suggestion: since `@boarding_status_override` is available at compile time, this is more efficient to implement with a macro. See https://github.com/mbta/concentrate/blob/master/lib/concentrate/parser/gtfs_realtime_enhanced.ex#L237-L267 later in this file.

paulswartz (changes requested):

lib/concentrate/group_filter/scheduled_stop_times.ex
Whoops: Just logging the duration, there was not a meaningful change in `group_filter` time with this change. Both with and without the change on my machine the `group_filter` stage took between 22-45ms with most cycles being around 30ms. I haven't charted and graphed them, but just skimming the distribution looks similar. I expect the various ets lookups done in this stage are probably what takes the most time but haven't done any research to very that.

Whoops: Another note: I'm actually OK with dropping this change all together. We discovered that at minimum dotcom is relying on some of these values, which is why. the other part of this PR is converting the status values back to their previous ones. I mainly left the change since the spec treats this as a free-form text field, and felt that being liberal in what we accept was probably best.

config/config.exs
Whoops: Removed

lib/concentrate/parser/gtfs_realtime_enhanced.ex
Whoops: Interesting! Until now I had assumed that doing pattern matching like that would more or less result in the compiler comparing against each string one by one to check for a match and would therefore be essentially linear time. I was going to write a little benchmark to see how that compared to the Map lookup, but it looks like others have already done it and pattern matching wins handily. I'll have to look more into how the compiler is optimizing these, in the meantime this has been updated.

bklebe: https://www.erlang.org/doc/efficiency_guide/binaryhandling.html is worth reading. Binary pattern matching in Erlang is optimized. 

The implementation of `Map.get/3` pattern matches on a map: https://github.com/elixir-lang/elixir/blob/c521bdb91a77b36be16fdf18d632ad7719de4f91/lib/elixir/lib/map.ex#L526-L537. I assume pattern matching on a map calls the same NIF as `:maps.get/3` does, but I can't figure out where exactly that's implemented because it's in native code.

PR #288, opened 2023-06-20
feat: Don't mix STU predictions and warn on time travel

Whoops:
#### Summary of changes

**Asana Ticket:** [🧠 🐛 Don't mix predictions: Multiple buses, same trip](https://app.asana.com/0/584764604969369/1204277338626221/f)

Reintroduces time travel filtering, with a fix to ensure we don't filter due to skipped stops.


Whoops:
I've updated this to contain the parts that I think still have merit, namely not mixing arrival/departure times when combining STUs and logging a warning (should it be error?) in the event of time travel.

lib/concentrate/stop_time_update.ex
```diff
@@ -35,13 +35,20 @@ defmodule Concentrate.StopTimeUpdate do
...
   defimpl Concentrate.Mergeable do
+    require Logger
+
     def key(%{trip_id: trip_id, stop_sequence: stop_sequence}), do: {trip_id, stop_sequence}
 
     def merge(first, second) do
+      time_stu =
+        if Concentrate.StopTimeUpdate.time(first) < Concentrate.StopTimeUpdate.time(second),
+          do: first,
+          else: second
```
bklebe, line 46: question: if these are equal we'll still choose the second. Is that the behavior we want?

Whoops: Honestly I think it's arbitrary. It's also not obvious that it's better to pick an optimistic prediction vs a pessimistic one. But I think that defaulting to `first` is probably the least surprising, so I've change it to `<=`.

lib/concentrate/group_filter/time_travel.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Concentrate.GroupFilter.TimeTravel do
+  @moduledoc """
+  Drops StopTimeUpdates that predict arriving at a later stop before departing an earlier one..
+  """
+  require Logger
+
+  @behaviour Concentrate.GroupFilter
```
bklebe, line 7: suggestion: sorry for the late suggestion, but what do you think about making this a reporter instead of a filter since it seems to be just logging certain characteristics? https://github.com/mbta/concentrate/tree/master/lib/concentrate/reporter. cc @paulswartz

bklebe (changes requested):
I think we should make sure this belongs as a filter and not a reporter before we merge — I pinged Paul to confirm but I believe it should be a reporter. Sorry for not catching this sooner. Hopefully the change isn't too big.

bklebe:
@Whoops sorry, just noticing Paul is out until the 23rd. Could you try to make this into a reporter and let me know if it doesn't make sense? Thanks.

lib/concentrate/group_filter/time_travel.ex
Whoops: Done. It's a bit funky since reporters are expected to maintain state, but no funkier than a GroupFilter that doesn't modify anything. Since the return values from Reporters will always be logged as `:info` I did the Logging within the reporter instead of returning KVs to be logged. The other option would be expecting (or allowing) the `log` function to return a log level, but that would involve modifying the existing `Reporters`.

bklebe: Hmmm. Without changing the other consumers, we could still make the default an `:info` log but allow passing in a level. This smells like the wrong abstraction to me though. What do you think about the reporter approach vs adding this as a check after we merge some STUs? Alternatively we could just downgrade this to an "info" but it does seem more like a warning. Curious what you think?

Whoops: Hmmm... the least invasive option I can think of would probably be adding a `level()` callback to the `Reporter` behavior, then using that when logging. Still requires modifying the `Consumer` and each `Reporter` implementations, but for the reporters at least it's just adding `def level, do: :info`.

That said another annoyance of returning KVs instead of handling the logging in the reporter is all the returned KVs will get [bundled up and logged in a single log entry](https://github.com/mbta/concentrate/blob/master/lib/concentrate/reporter/consumer.ex#L29), which in the event of multiple time travels will be confusing at best. Which I think talks me out of it. 

When you suggest adding it as a post-merge check are you suggesting adding a validation step [here](https://github.com/mbta/concentrate/blob/master/lib/concentrate/merge_filter.ex#L124)? That might be cleanest since it doesn't seem like any of the existing abstractions map well to what we're trying to do. On the other hand I could also probably be talked into the idea that this is to annoying to implement for the value we're going to get out of it, and that we should add some kind of alerting to the validators outside of concentrate instead (e.x. can we dream up something that alerts us if Google reports and error in our realtime feed). That might be more bang for the buck on effort, since this is only one specific check.

lib/concentrate/reporter/time_travel.ex
```diff
@@ -0,0 +1,46 @@
+defmodule Concentrate.Reporter.TimeTravel do
+  @moduledoc """
+  Drops StopTimeUpdates that predict arriving at a later stop before departing an earlier one..
```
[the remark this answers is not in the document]
Whoops: No, I'm pretty sure we decided we didn't want to touch the feed, that trying to modify bad data into good data was to error prone. I just forgot to update the moduledoc when repurposing the code.

lib/concentrate/group_filter/time_travel.ex
Whoops: Pondering this a few days I think sticking with the reporter is the right thing for now at least. It is inline with what reporters are meant to do, and I don't think it's worth building out a new abstraction for this one feature. If we start adding more post-processing validators we can either adjust how reporters work and create a new abstraction.

lib/concentrate/reporter/time_travel.ex
```diff
@@ -0,0 +1,46 @@
...
+  """
+  require Logger
+
+  @behaviour Concentrate.Reporter
+  @impl Concentrate.Reporter
+
+  def init do
```
bklebe, line 10:
```suggestion
  @impl Concentrate.Reporter
  def init do
```

lib/concentrate/reporter/time_travel.ex
```diff
@@ -0,0 +1,46 @@
...
+    []
+  end
```
bklebe, line 12:
```suggestion
  @behaviour Concentrate.Reporter
  
  @impl Concentrate.Reporter
  def init do
    []
  end
```

lib/concentrate/reporter/time_travel.ex
```diff
@@ -0,0 +1,46 @@
...
+
+  @impl Concentrate.Reporter
+  def log(groups, _) do
+    Enum.each(groups, &log_group/1)
+    {[], []}
+  end
+
+  def log_group({_td, _vps, stop_time_updates}) do
```
bklebe, line 20:
```suggestion
  defp log_group({_td, _vps, stop_time_updates}) do
```

lib/concentrate/reporter/time_travel.ex
```diff
@@ -0,0 +1,46 @@
...
+    second_time = second.arrival_time || second.departure_time
+
+    if not is_nil(first_time) and not is_nil(second_time) and second_time < first_time do
+      Logger.warning(
+        "event=time_travel trip_id=#{first.trip_id} first_stop=#{first.stop_sequence} first_time=#{first_time} second_stop=#{second.stop_sequence} second_time=#{second_time}"
+      )
+    end
+
+    time_travel_check([second | rest])
+  end
```
bklebe, line 45:
```suggestion
  defp time_travel_check([first, second | rest]) do
    first_time = first.departure_time || first.arrival_time
    second_time = second.arrival_time || second.departure_time

    if not is_nil(first_time) and not is_nil(second_time) and second_time < first_time do
      Logger.warning(
        "event=time_travel trip_id=#{first.trip_id} first_stop=#{first.stop_sequence} first_time=#{first_time} second_stop=#{second.stop_sequence} second_time=#{second_time}"
      )
    end

    time_travel_check([second | rest])
  end
  
  defp time_travel_check(list) when is_list(list) do
    nil
  end
```

bklebe, line 45: question: do you think it makes sense to put this last case of `time_travel_check` first and then condense the first two into a case that checks if the argument is a list and returns nil?

bklebe (changes requested):
Sorry for the late review here. The minor formatting and function privacy issues need to be addressed but this looks good overall. I think `time_travel_check` could be simplified as well.

lib/concentrate/reporter/time_travel.ex
Whoops: Yup!

Whoops:
I've addressed the comments and simplified the if on the `time_travel_check`.

bklebe (approved):

PR #289, opened 2023-06-30
Update to cancelled_trip group filter 

ErinLMoore:
#### Summary of changes

Cancelled trip group filter now filters out trips where all stop updates have SKIPPED status

**Asana Ticket:** [Add trip-level schedule_relationship for canceled bus trips](https://app.asana.com/0/505721188639414/1204778267687026/f)

This change surfaces a trip-level schedule relationship of CANCELED for bus trips where all of the associated stop time updates have been given the SKIPPED status.

### DIFF
--- a/lib/concentrate/group_filter/cancelled_trip.ex
+++ b/lib/concentrate/group_filter/cancelled_trip.ex
@@ -18,11 +18,11 @@ defmodule Concentrate.GroupFilter.CancelledTrip do
       TripDescriptor.schedule_relationship(td) == :CANCELED ->
         cancel_group(group)
 
+      Enum.all?(stop_time_updates, &StopTimeUpdate.skipped?(&1)) ->
+        cancel_group(group)
+
       is_nil(time) ->
         group
-
-      Enum.all?(stop_time_updates, &(StopTimeUpdate.schedule_relationship(&1) == :SKIPPED)) ->
-        cancel_group(group)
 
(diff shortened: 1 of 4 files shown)


paulswartz (commented):
could you deploy this to one of the dev environments so we can check out how it works?

ErinLMoore:
> could you deploy this to one of the dev environments so we can check out how it works?

I just deployed it on `dev`

lib/concentrate/group_filter/cancelled_trip.ex
```diff
@@ -27,6 +27,9 @@ defmodule Concentrate.GroupFilter.CancelledTrip do
       is_binary(route_id) and module.route_cancelled?(route_id, time) ->
         cancel_group(group)
 
+      Enum.all?(stop_time_updates, &StopTimeUpdate.skipped?(&1)) ->
+        cancel_group(group)
```
bklebe, line 31: question: do you know if we should also cancel the group if the STUs are all skipped but the first one's time is `nil` like the case on line 21?

ErinLMoore: I'm actually not sure which rule should take precedence, and there doesn't seem to be a test around the case on line 21.  I'll ask on the ticket and see if someone can clarify.

bklebe (approved):

PR #294, opened 2023-07-25
fix: filter stop time updates without arrival/departure

Whoops:
#### Summary of changes

**Asana Ticket:** [🐛 API publishing bus predictions with no times or schedule relationship](https://app.asana.com/0/584764604969369/1205057459791776/f)

This filters out StopTimeUpdates which are scheduled, but have no arrival/departure. These occur when the BusLoc feed contains stops which Swiftly either isn't yet predicting, or is no longer publishing predictions for because the bus is already well past that stop.


Whoops:
> Instead, I'd look at filtering these out of non-enhanced outputs, as those wouldn't include the `boarding_status` field.

Ugh, and in fact they already are being filtered from non-enchanced feeds (https://github.com/mbta/concentrate/blob/master/lib/concentrate/encoder/trip_updates.ex#L28). But the predictions API uses the enhanced feed and so is publishing a bunch of nonsensical bus predictions. I'll need to come up with a new idea to deal with those. Especially since apparently similar predictions are valid in other contexts.

paulswartz:
Given the original report, it looks like the issue is starting with Busloc. That application should never be generating predictions without at least one of an arrival/departure time AFAIK.

Whoops:
As far as I can tell, it'll never output an arrival/departure time. Some waivers it publishes as skipped, some as scheduled (https://github.com/mbta/busloc/blob/master/lib/busloc/waiver/encoder/trip_updates_enhanced.ex#L71). The last time we went down this rabbit-hole we were unable to get a clear answer as to the purpose of the scheduled updates.

Whoops:
That seems reasonable. I'll make the change. Thanks!

lib/concentrate/filter/unscheduled_scheduled_stop.ex
```diff
@@ -0,0 +1,25 @@
...
+  """
+  @behaviour Concentrate.Filter
+  alias Concentrate.StopTimeUpdate
+  require Logger
+
+  @impl Concentrate.Filter
+  def filter(%StopTimeUpdate{} = stu) do
+    scheduled? =
+      StopTimeUpdate.schedule_relationship(stu) == :SCHEDULED ||
+        StopTimeUpdate.schedule_relationship(stu) == nil
```
paulswartz, line 13: question: is it possible for this to be `nil`? I thought the parser converted it to the default value of `:SCHEDULED`?

Whoops: It looks like your correct, I've removed the condition. I was working off the spec and the fact that concentrate will omit `schedule_relationship` in it's output if it's `:SCHEDULED`

paulswartz (approved):
🍰 

code looks good! can you deploy again to make sure it's working okay?

Whoops:
> code looks good! can you deploy again to make sure it's working okay?

Yep! I sent it to dev-green, I'll keep an eye on it for issues.

PR #304, opened 2023-09-15
feat: parsed DIFFERENTIAL feeds coming in from MQTT

paulswartz:
#### Summary of changes

**Asana Ticket:** [accept DIFFERENTIAL feeds from Busloc](https://app.asana.com/0/1205365039348894/1205499829140889/f)

mbta/busloc#324 adds support to publish DIFFERENTIAL updates. This PR (which should go first) accepts those updates.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -2,7 +2,7 @@
 # and its dependencies with the aid of the Mix.Config module.
 import Config
 
-config :logger, level: :debug
+config :logger, level: :info
 
 config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -80,6 +80,7 @@ config :concentrate,
     Concentrate.GroupFilter.TripDescriptorTimestamp
   ],
   source_reporters: [
+    Concentrate.SourceReporter.Basic,
     Concentrate.SourceReporter.Latency
   ],
   reporters: [
--- a/lib/concentrate/trip_descriptor.ex
+++ b/lib/concentrate/trip_descriptor.ex
@@ -16,6 +16,18 @@ defmodule Concentrate.TripDescriptor do
     schedule_relationship: :SCHEDULED
   ])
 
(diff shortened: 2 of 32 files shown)


config/config.exs
```diff
@@ -2,7 +2,7 @@
 # and its dependencies with the aid of the Mix.Config module.
 import Config
 
-config :logger, level: :debug
+config :logger, level: :info
```
nlwstein, line 5: Is this an intentional change?

lib/concentrate/trip_descriptor.ex
```diff
@@ -16,6 +16,18 @@ defmodule Concentrate.TripDescriptor do
     schedule_relationship: :SCHEDULED
   ])
 
+  def timestamp_truncated(%__MODULE__{timestamp: number}) when is_integer(number) do
```
nlwstein, line 19: Is it worth adding a condition to handle if `number` is valid but cast to a string?

nlwstein (changes requested):
The vast majority of this looks good to me! 

I only had a q regarding the expectation that timestamp will always be float or int. I suspect an accidental string would be more likely than a float value for an epoch timestamp, but it's not a blocking change either.

config/config.exs
paulswartz: Yep, it's meant to clean up the very aggressive logs from Emqtt: https://github.com/mbta/concentrate/pull/304/commits/431481213f108b1148bab2a6cb8457c92fc7ea96

lib/concentrate/trip_descriptor.ex
paulswartz: Good question! At the moment, it's an epoch time, either  integer, or a float (with a number of millisconds/microseconds). Given that only one producer generates floats (Busloc) I don't think it's worth extra complexity. 

If we did want to add the extra complexity, we'd probably add it at the parsing stage, so either way it wouldn't go here.

nlwstein (approved):

PR #305, opened 2023-09-15
feat: allow integer vehicle IDs

bklebe:
#### Summary of changes

**Asana Ticket:** [allow integer vehicle IDs in concentrate in case they get passed through](https://app.asana.com/0/0/1205504059419471/f)

Logs a warning if an integer vehicle ID is encountered and stringifies it with `Integer.to_string/1`.


paulswartz (commented):
question: should this also handle the `id` field in the VehiclePosition struct?

PR #323, opened 2023-11-03
fix: don't include non-revenue trips in feeds

Whoops:
#### Summary of changes

**Asana Ticket:** [🧠🚫💰 Surface non-revenue train information through our public-facing data feeds/API](https://app.asana.com/0/505721188639414/1205761861495473/f)

Note: We will shortly want to start exposing non-revenue trips in the enhanced feed. This implementation is chosen to make that update trivial.

### DIFF
--- a/lib/concentrate/encoder/gtfs_realtime_helpers.ex
+++ b/lib/concentrate/encoder/gtfs_realtime_helpers.ex
@@ -140,6 +140,16 @@ defmodule Concentrate.Encoder.GTFSRealtimeHelpers do
   def schedule_relationship(:SCHEDULED), do: nil
   def schedule_relationship(relationship), do: relationship
 
+  @doc """
+  Returns true if the group is non-revenue
+  """
+  def is_non_revenue?({td, _, stus} = _group) do
+    td &&
+      td.trip_id &&
+      String.starts_with?(td.trip_id, "NONREV") &&
+      Enum.all?(stus, &(&1.schedule_relationship == :SKIPPED))
+  end
+
   defp group_by_trip_id(%TripDescriptor{} = td, map) do
     if trip_id = TripDescriptor.trip_id(td) do
       Map.update(map, trip_id, {td, [], []}, &add_trip_descriptor(&1, td))
--- a/lib/concentrate/encoder/trip_updates_enhanced.ex
+++ b/lib/concentrate/encoder/trip_updates_enhanced.ex
@@ -8,6 +8,8 @@ defmodule Concentrate.Encoder.TripUpdatesEnhanced do
 
   @impl Concentrate.Encoder
   def encode_groups(groups, opts \\ []) when is_list(groups) do
+    groups = Enum.reject(groups, &is_non_revenue?/1)
+
     message = %{
       header: feed_header(opts),
       entity: trip_update_feed_entity(groups, &build_stop_time_update/1, &enhanced_data/1)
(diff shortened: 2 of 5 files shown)


lib/concentrate/encoder/gtfs_realtime_helpers.ex
```diff
@@ -140,6 +140,16 @@ defmodule Concentrate.Encoder.GTFSRealtimeHelpers do
   def schedule_relationship(:SCHEDULED), do: nil
   def schedule_relationship(relationship), do: relationship
 
+  @doc """
+  Returns true if the group is non-revenue
+  """
+  def is_non_revenue?({td, _, stus} = _group) do
```
paulswartz, line 146: suggestion(nitpick):
```suggestion
  def non_revenue?({td, _, stus} = _group) do
```

question: I thought the plan was to provide a flag in the output to indicate that a trip was non-revenue, rather than parsing the trip IDs?

Whoops: There's discussion around the flag happening in Slack currently, but given that this fixes us sharing information we currently say we aren't, I don't think this PR should be held up on a final direction there.

Whoops:
This has been updated to use the (soon to be) newly added revenue field in RTR rather than filtering on `trip_id`.

lib/concentrate/encoder/trip_updates_enhanced.ex
```diff
@@ -17,7 +19,10 @@ defmodule Concentrate.Encoder.TripUpdatesEnhanced do
   end
 
   defp enhanced_data(update) do
-    %{route_pattern_id: TripDescriptor.route_pattern_id(update)}
+    %{
+      route_pattern_id: TripDescriptor.route_pattern_id(update),
+      revenue: TripDescriptor.revenue(update)
```
paulswartz, line 24: suggestion: given that `true` is the default, what about only including it when the value is `false`?

lib/concentrate/encoder/vehicle_positions_enhanced.ex
```diff
@@ -19,7 +21,7 @@ defmodule Concentrate.Encoder.VehiclePositionsEnhanced do
   end
 
   def build_entity({%TripDescriptor{} = td, vps, _stus}) do
-    trip = trip_descriptor(td)
+    trip = trip_descriptor(td) |> Map.put("revenue", TripDescriptor.revenue(td))
```
paulswartz, line 24:
```suggestion
    trip = td
    |> trip_descriptor() 
    |> Map.put("revenue", TripDescriptor.revenue(td))
```

lib/concentrate/encoder/trip_updates_enhanced.ex
Whoops: If you feel strongly about it, I can. My main hesitation is that in `nil` ~ falsey languages, the naive interpretation is the opposite of what we actually mean. I'd tend to argue there shouldn't be a default, but maintaining current behavior on existing feeds kind of forces our hand there.

lib/concentrate/encoder/vehicle_positions_enhanced.ex
Whoops: We might want to consider turning on Credo's `PipeChainStart` check. It's explicitly off at the moment.

paulswartz: It wasn't originally: https://github.com/mbta/concentrate/commit/70321490129465324fd8e18faa3b4f5b9b938f17

It looks like the default changed when we re-built the `.credo.exs`: https://github.com/mbta/concentrate/commit/c51fabfb969192ad1e6395475611f0649aa86168

lib/concentrate/encoder/trip_updates_enhanced.ex
paulswartz: As you note, existing clients which care about this value will need to treat a missing value as `true` (as it was true for all trips before). But I don't think it needs to block the PR.

paulswartz (approved):

PR #325, opened 2023-11-15
fix: enforce schema for multi_carriage_details for protobuf export

nlwstein:
#### Summary of changes

This branch adds a bunch of validation to avoid values that the Protobuf encoder does not like - specifically `nil` in place of the `occupancy_percentage` and `id` fields that is common throughout our RTR feed - and enforce the default values. 

The changes to the default `config.exs` that I used: 

```
diff --git a/config/config.exs b/config/config.exs
index ea5b6ff..aa39f48 100644
--- a/config/config.exs
+++ b/config/config.exs
@@ -38,6 +38,10 @@ config :concentrate,
     gtfs_realtime: [
       vehicle_positions: "https://cdn.mbta.com/realtime/VehiclePositions.pb",
       trip_updates: "https://cdn.mbta.com/realtime/TripUpdates.pb"
+    ],
+    "gtfs_realtime_enhanced": [
+      rtr_tripupdates: "https://mbta-gtfs-s3.s3.amazonaws.com/rtr/TripUpdates_enhanced.json",
+      rtr_vehiclepositions: "https://mbta-gtfs-s3.s3.amazonaws.com/rtr/VehiclePositions_enhanced.json"
     ]
   ],
   alerts: [
@@ -100,7 +104,7 @@ config :concentrate,
     ]
   ],
   sinks: [
-    filesystem: [directory: "/tmp"]
+    filesystem: [directory: "output/"]
   ],
   file_tap: [
     enabled?: false
```

This was validated against by using https://github.com/beatlabs/proton to convert the resultant pb file into a JSON file for inspection, like this: `proton json -f ./src/gtfs-realtime.proto output/VehiclePositions.pb > result.json`

Example output: 

```
{
      "id": "O-54798EDB",
      "vehicle": {
        "trip": {
          "tripId": "ADDED-1581698495",
          "startTime": "15:17:05",
          "startDate": "20231115",
          "scheduleRelationship": "ADDED",
          "routeId": "Orange",
          "directionId": 1
        },
        "position": {
          "latitude": 42.36473,
          "longitude": -71.05958,
          "bearing": 330
        },
        "currentStopSequence": 130,
        "currentStatus": "INCOMING_AT",
        "timestamp": "1700081065",
        "stopId": "70027",
        "vehicle": { "id": "O-54798EDB", "label": "1468" },
        "multiCarriageDetails": [
          {
            "id": "",
            "label": "1468",
            "occupancyStatus": "STANDING_ROOM_ONLY",
            "occupancyPercentage": 39,
            "carriageSequence": 1
          },
          {
            "id": "",
            "label": "1469",
            "occupancyStatus": "STANDING_ROOM_ONLY",
            "occupancyPercentage": 40,
            "carriageSequence": 2
          },
          {
            "id": "",
            "label": "1422",
            "occupancyStatus": "STANDING_ROOM_ONLY",
            "occupancyPercentage": 40,
            "carriageSequence": 3
          },
          {
            "id": "",
            "label": "1423",
            "occupancyStatus": "STANDING_ROOM_ONLY",
            "occupancyPercentage": 31,
            "carriageSequence": 4
          },
          {
            "id": "",
            "label": "1471",
            "occupancyStatus": "STANDING_ROOM_ONLY",
            "occupancyPercentage": 29,
            "carriageSequence": 5
          },
          {
            "id": "",
            "label": "1470",
            "occupancyStatus": "STANDING_ROOM_ONLY",
            "occupancyPercentage": 35,
            "carriageSequence": 6
          }
        ]
      }
    }
```

**Asana Ticket:** [🐛 🧠 MultiCarriageDetails in VehiclePositions PB file](https://app.asana.com/0/505721188639414/1205920863572433/f)


nlwstein:
Idea: Would it be worth having an integration test that runs on a schedule that pulls our production (and other envs?) Protobuf and JSON payloads, and diff them to ensure they are the same?

nlwstein:
> Idea: Would it be worth having an integration test that runs on a schedule that pulls our production (and other envs?) Protobuf and JSON payloads, and diff them to ensure they are the same?

This could be a GitHub Action that runs on a schedule as part of this repo (or another) - @bklebe @meagharty this might be an interesting integration test task 😄

paulswartz:
> This branch adds a bunch of validation to avoid values that the Protobuf encoder does not like - specifically nil in place of the occupancy_percentage and id fields that is common throughout our RTR feed - and enforce the default values.

Can you say more about this? `id` and `occupancy_percentage` are both in the `.proto` file as optional, so they should be able to be not set in the output.

paulswartz:
> Idea: Would it be worth having an integration test that runs on a schedule that pulls our production (and other envs?) Protobuf and JSON payloads, and diff them to ensure they are the same?

I thought the issue was that the Enhanced JSON and Protobuf didn't have the same fields, not the regular JSON?

nlwstein:
> I thought the issue was that the Enhanced JSON and Protobuf didn't have the same fields, not the regular JSON?

I know that for `multi_carriage_details` I had to add it to the build for the Protobuf and JSON Enhanced feeds separately, so it is definitely possible that one or the other could be missing fields. Are you referring to fields like `consist`, which aren't actually part of the spec? Maybe we could use an ignore list to skip over the fields that aren't shared.  

> Can you say more about this? `id` and `occupancy_percentage` are both in the `.proto` file as optional, so they should be able to be not set in the output.

RTR sets them to `nil` explicitly in a few cases, which Protobuf doesn't understand. You're right that they are optional per spec so I changed it to remove them instead of setting default values.

lib/concentrate/vehicle_position/carriage_details.ex
```diff
@@ -0,0 +1,62 @@
...
+          do: atomized_carriage_details.carriage_sequence,
+          else: nil
+        ),
+      occupancy_status:
+        if(is_atom(unatomized_occupancy_status),
+          do: unatomized_occupancy_status,
+          else: String.to_atom(unatomized_occupancy_status)
+        ),
+      occupancy_percentage:
+        if(Map.has_key?(atomized_carriage_details, :occupancy_percentage),
```
paulswartz, line 56: question: is this different from `Map.get(atomized_carriage_details, :occupancy_percentage)`?

lib/concentrate/vehicle_position/carriage_details.ex
```diff
@@ -0,0 +1,62 @@
...
+          do: atomized_carriage_details.occupancy_percentage,
+          else: nil
+        )
```
paulswartz, line 59:
```suggestion
    Map.take(atomized_carriage_details, ~w(id label carriage_sequence occupancy_status occupancy_percentage)a)
```

test/concentrate/encoder/vehicle_positions_test.exs
```diff
@@ -152,6 +152,37 @@ defmodule Concentrate.Encoder.VehiclePositionsTest do
...
+          multi_carriage_details: [
+            %{
+              id: "",
+              label: "3022",
+              carriage_sequence: 1,
+              occupancy_status: :MANY_SEATS_AVAILABLE,
+              occupancy_percentage: 10
+            },
+            %{
+              id: "",
```
paulswartz, line 173: question: what happens with the `id` if it's not present in the initial data?

paulswartz (commented):
question: should `VehiclePositionsEnhanced` be sharing this code to ensure that the PB and Enhanced JSON match?

test/concentrate/encoder/vehicle_positions_test.exs
nlwstein: The test still passes w/ the missing id. The code was changed to simply drop fields that are nil.

lib/concentrate/vehicle_position/carriage_details.ex
```diff
@@ -0,0 +1,36 @@
...
+  # Convert to atomized keys, so that both atoms and string keys are supported:
+  defp get_atomized_carriage_details(carriage_details) do
+    atomized_carriage_details =
+      for {key, val} <- carriage_details,
+          into: %{},
+          do:
+            {if(is_atom(key), do: key, else: String.to_atom(key)),
+             if(key in ["occupancy_status", :occupancy_status], do: String.to_atom(val), else: val)}
+
+    Map.take(
```
nlwstein, line 31: @paulswartz I kept your suggested change despite the revert - it's very clean, I just had to update the logic prior to cast the `occupancy_status` *value* into an atom. That was taking place in that mostly copy paste section this eliminated 😄

nlwstein:
> question: should `VehiclePositionsEnhanced` be sharing this code to ensure that the PB and Enhanced JSON match?

There's no harm in adding it there too, updated!

paulswartz (commented):
question: can we throw this up on dev-green to make sure the performance is fine?

question: apologies for only thinking about this now, but: should we be handling this issue at parse time, rather than render time? Right now, we're handling the question of "does multi_carriage_details have the right fields?" when we write the fields back, but for all other fields, we read them at parse time and then the encoding step is basically a no-op. should we be doing that here as well?

nlwstein:
> question: can we throw this up on dev-green to make sure the performance is fine?

No problem w/ me to put it up on dev-green, I'll double check at standup. 

> question: apologies for only thinking about this now, but: should we be handling this issue at parse time, rather than render time? Right now, we're handling the question of "does multi_carriage_details have the right fields?" when we write the fields back, but for all other fields, we read them at parse time and then the encoding step is basically a no-op. should we be doing that here as well?

I'm mixed here because the issue is specifically related to encoding. It probably would be trivial to move the logic up the pipeline, though!

lib/concentrate/vehicle_position/carriage_details.ex
```diff
@@ -0,0 +1,39 @@
...
+    end)
+  end
+
+  # Convert to atomized keys, so that both atoms and string keys are supported:
+  defp get_atomized_carriage_details(carriage_details) do
+    atomized_carriage_details =
+      for {key, val} <- carriage_details,
+          into: %{},
+          do:
+            {if(is_atom(key), do: key, else: String.to_atom(key)),
```
paulswartz, line 28: suggestion!: this isn't safe for arbitrary inputs, as it can allow for atom exhaustion: https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/atom_exhaustion.html

That link has a few different options for how to handle it.

lib/concentrate/vehicle_position/carriage_details.ex
```diff
@@ -0,0 +1,39 @@
...
+             if(key in ["occupancy_status", :occupancy_status] and not is_atom(val),
+               do: String.to_atom(val),
```
paulswartz, line 30: See above on atom exhaustion.

lib/concentrate/vehicle_position/carriage_details.ex
nlwstein: > suggestion!: this isn't safe for arbitrary inputs, as it can allow for atom exhaustion: https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/atom_exhaustion.html
> 
> That link has a few different options for how to handle it.

I have updated to use String.to_existing_atom and added some code mostly to explicitly declare the atoms we need for this use case.

paulswartz (approved):

PR #346, opened 2024-03-21
🧠 Implement "uncertainty adapter" within Concentrate

bfauble:
#### Summary of changes

**Asana Ticket:** [🧠 Implement "uncertainty adapter" within Concentrate](https://app.asana.com/0/505721188639414/1206586251381038/f)

Update to handle `update_type` field being present at the `TripUpdate` level from `TripUpdates_enhanced.json` feed from RTR. 

`"mid_trip" == 60`
`"at_terminal" == 120`
`"reverse_trip" == 360`

```
%{
  "trip" => %{
    "trip_id" => "trip",
    "route_id" => "route"
  },
  "update_type" => "mid_trip",
  "stop_time_update" => [%{
    "arrival" => %{"time" => 100, "uncertainty" => 60},
    "departure" => %{"time" => 200, "uncertainty" => 60}
  }]
}
```

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206695607194179

### DIFF
--- a/lib/concentrate/group_filter/uncertainty_value.ex
+++ b/lib/concentrate/group_filter/uncertainty_value.ex
@@ -7,7 +7,7 @@ defmodule Concentrate.GroupFilter.UncertaintyValue do
 
   @impl Concentrate.GroupFilter
   def filter({%TripDescriptor{update_type: update_type} = td, vps, stus})
-      when is_binary(update_type) do
+      when not is_nil(update_type) do
     stus =
       update_type
       |> calculate_uncertainty()


paulswartz (commented):
question: is the parser the right place to be making a change to the data? how is this going to work in the future when we want to incorporate other sources of data into these uncertainty updates?

I'm also a little confused about why this particular change is happening in Concentrate, vs RTR. Since RTR is putting in the `update_type` value, why isn't it also updating the `uncertainty` values in the feed it puts out?

bklebe (commented):
@bfauble Thanks for this change! A couple questions:

- I'm a little confused about the implementation plan here. Why are we adding `update_type` to RTR only to convert it back into an opaque uncertainty value in Concentrate?
- Do you think this would be better as a `GroupFilter` than a parser extension? That's what I was imagining, though seeing as the implementation plan here is also different from what I expected, maybe putting it in the parser makes sense.

Let's discuss at standup :)

bklebe (approved):

bklebe (changes requested):
As discussed in Slack, this needs updated now that RTR is providing the `update_type` field.

paulswartz (commented):
suggestion: given this rationale for the ticket, I don't think this can live in the parser.
> [...] this would allow [...] for Concentrate to source data from multiple sources to assign and adjust uncertainty values.
> - Will be able to ingest multiple types of inputs to make adjustments to uncertainty in the future - this will only be used to make uncertainty changes based on OIO-indicated sign headway mode initially, but we'll be looking to make more substantial use of this in the future.

If it lives in the parser, there's no way for it to consume multiple sources of data.

paulswartz:
@bklebe [mentioned](https://github.com/mbta/concentrate/pull/346#pullrequestreview-1957810700) using a `GroupFilter`, which seems like the appropriate place to me as well.

bklebe (approved):

paulswartz:
@bfauble I don't see the `update_type` value in the dev Concentrate feed after the deploy.

lib/concentrate/group_filter/uncertainty_value.ex
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def filter(other), do: other
+
+  defp set_uncertainty(nil, stus), do: stus
+
+  defp set_uncertainty(update_type, stus) do
+    Enum.map(stus, fn stu ->
+      StopTimeUpdate.update_uncertainty(stu, calculate_uncertainty(update_type))
+    end)
```
paulswartz, line 22: suggestion: since `calculate_uncertainty` only depends on the `update_type` you shouldn't calculate it within the `map`.

lib/concentrate/group_filter/uncertainty_value.ex
```diff
@@ -0,0 +1,29 @@
...
+    {td, vps, stus}
+  end
+
+  def filter(other), do: other
+
+  defp set_uncertainty(nil, stus), do: stus
+
+  defp set_uncertainty(update_type, stus) do
+    Enum.map(stus, fn stu ->
+      StopTimeUpdate.update_uncertainty(stu, calculate_uncertainty(update_type))
```
paulswartz, line 21: question: this looks like it always sets the `uncertainty` even if `update_type` isn't set. won't this override the uncertainty we get from non-RTR sources?

lib/concentrate/trip_descriptor.ex
```diff
@@ -59,13 +60,17 @@ defmodule Concentrate.TripDescriptor do
           vehicle_id: first.vehicle_id || second.vehicle_id,
           timestamp: first.timestamp || second.timestamp,
           last_trip: first.last_trip || second.last_trip,
-          schedule_relationship: merge_schedule_relationship(first, second)
+          schedule_relationship: merge_schedule_relationship(first, second),
+          update_type: merge_update_type(first.update_type, second.update_type)
```
paulswartz, line 64: suggestion: `first.update_type || second.update_type`

test/concentrate/group_filter/unceratinty_value_test.exs
```diff
@@ -0,0 +1,73 @@
...
+      ]
+
+      {^td, [], processed_stus} = filter({td, [], stus})
+
+      assert Enum.all?(processed_stus, fn procced_stu ->
+               StopTimeUpdate.uncertainty(procced_stu) == 360
+             end)
+    end
+
+    test "populates uncertainty in TripDescriptor based on update_type value of other" do
```
paulswartz, line 57: suggestion: can you add a test than an `uncertainty` value without an `update_type` is kept?

config/config.exs
```diff
@@ -73,7 +73,8 @@ config :concentrate,
     Concentrate.GroupFilter.VehicleAtSkippedStop,
     Concentrate.GroupFilter.VehicleStopMatch,
     Concentrate.GroupFilter.SkippedStopOnAddedTrip,
-    Concentrate.GroupFilter.TripDescriptorTimestamp
+    Concentrate.GroupFilter.TripDescriptorTimestamp,
+    Concentrate.GropuFilter.UncertaintyValue
```
paulswartz, line 77: suggestion: typo, should be `Concentrate.GroupFilter.UncertaintyValue`

lib/concentrate/encoder/gtfs_realtime_helpers.ex
```diff
@@ -221,7 +223,8 @@ defmodule Concentrate.Encoder.GTFSRealtimeHelpers do
                 trip: trip,
                 stop_time_update: stop_time_update,
                 vehicle: vehicle,
-                timestamp: timestamp
+                timestamp: timestamp,
+                update_type: update_type
```
paulswartz, line 227: question: do you need the changes to this file? This are generic to both enhanced and non-enhanced feeds, and the non-enhanced feeds don't have an `update_type`.

paulswartz (changes requested):

paulswartz (commented):
Definitely looks like there's a performance drop with this: 
Screenshot 2024-04-25 at 9 03 41 AM
Let me know if you need help looking into it!

paulswartz (approved):
Performance looks better now!

PR #349, opened 2024-04-30
Filter stus from updates based on stops currently flagged for suppression

bfauble:
#### Summary of changes

**Asana Ticket:** [Add filter of stus based on config state](Add filter of stus based on config state)

* Filters stus based on the `route_id`, `direction_id`, and `stop_id` stored in `StopPredictionStatus` state.
* TODO: Fetch and store stop prediction status from signs config in followup PR.

### DIFF
--- a/lib/concentrate/group_filter/suppress_stop_time_update.ex
+++ b/lib/concentrate/group_filter/suppress_stop_time_update.ex
@@ -26,7 +26,7 @@ defmodule Concentrate.GroupFilter.SuppressStopTimeUpdate do
             )
           end)
 
-        _ = log_suppressed_stus(suppressed_stus, route_id, direction_id)
+        log_suppressed_stus(suppressed_stus, route_id, direction_id)
 
         {td, vps, unsuppressed_stus}
     end
@@ -38,7 +38,7 @@ defmodule Concentrate.GroupFilter.SuppressStopTimeUpdate do
     do: MapSet.member?(suppressed_stops, stop_id)
 
   defp log_suppressed_stus(stus, route_id, direction_id) do
-    Enum.map(stus, fn stu ->
+    Enum.each(stus, fn stu ->
       stop_id = StopTimeUpdate.stop_id(stu)
 
       Logger.info(fn ->


lib/concentrate/group_filter/supress_stop_time_update.ex
```diff
@@ -0,0 +1,48 @@
...
+
+  @impl Concentrate.GroupFilter
+  def filter({td, vps, stus}) do
+    case StopPredictionStatus.flagged_stops_on_route(td) do
+      nil ->
+        {td, vps, stus}
+
+      supressed_stops ->
+        route_id = TripDescriptor.route_id(td)
+        direction_id = TripDescriptor.direction_id(td)
```
bklebe, line 19: suggestion: since these are also required by `flagged_stops_on_route`, which also doesn't need the whole TripDescriptor, perhaps consider pulling them up to the top level of `filter`, and refactoring `flagged_stops_on_route` to consume `route_id` and `direction_id` directly? It would make that function simpler and I think it would also make the connection between its use of route and direction and `filter`'s use of it here more explicit.

lib/concentrate/group_filter/supress_stop_time_update.ex
```diff
@@ -0,0 +1,48 @@
...
+
+        unsupressed_stus =
+          Enum.reject(stus, fn stu ->
+            stop_id_supressed?(
+              supressed_stops,
+              StopTimeUpdate.stop_id(stu),
+              route_id,
+              direction_id
+            )
+          end)
```
bklebe, line 29: suggestion(non-blocking): if we converted this to `Enum.split_with` instead of reject, we could pull the logging side effect up a level out of `stop_id_suppressed?` (sp) (perhaps replacing with `Enum.each`) and remove the `route_id` and `direction_id` arguments. I guess this is sort of a stylistic difference in that I prefer to hoist side effects up as high as possible. What do you think?

bklebe (changes requested):

lib/concentrate/group_filter/suppress_stop_time_update.ex
```diff
@@ -0,0 +1,49 @@
...
+    end
+  end
+
+  def filter(other), do: other
+
+  defp stop_id_suppressed?(suppressed_stops, stop_id),
+    do: MapSet.member?(suppressed_stops, stop_id)
+
+  defp log_suppressed_stus(stus, route_id, direction_id) do
+    Enum.map(stus, fn stu ->
```
bklebe, line 41: suggestion: use `Enum.each` instead of `Enum.map` since we're only interested in the side effect here

bklebe (approved):

PR #350, opened 2024-05-02
Parse signs config and store info in state to be used in group filter

bfauble:
#### Summary of changes

**Asana Ticket:** [Parse config and store info in state](https://app.asana.com/0/0/1207186593091892/f)

* Moves state logic into correct namespace
* Parser for parsing signs config feed
* Group filter uses data from state
* TODO: Supervisor to do the actual fetching of feed to feed into state


nlwstein:
This might be better-suited to a quick discussion / call, but could we deploy to a dev environment and validate this behavior manually? There's a lot going on here logic-wise, and I think if we both smoke tested it in dev that would be confidence-inducing. 

The code itself looks sound syntactically 🙂

nlwstein (approved):
OK - that makes sense - I figure we can review this more conclusively once it's enabled in dev @bfauble. The code otherwise looks syntactically good so I think this is good to merge.

PR #351, opened 2024-05-08
🧠 Ingest OIO trigger and suppress relevant predictions 

bfauble:
#### Summary of changes

**Asana Ticket:** [Fetch Config](https://app.asana.com/0/0/1207226872843767/f)

* Add actual fetching from live feeds
* NOTE: updating field placement in json blob config in devops pr here: https://github.com/mbta/devops/pull/1769
* Will need to wait on devops merge/apply to be able to properly make use of config

### DIFF
--- a/test/concentrate/filter/suppress/stop_prediction_status_test.exs
+++ b/test/concentrate/filter/suppress/stop_prediction_status_test.exs
@@ -20,6 +20,9 @@ defmodule Concentrate.Filter.Suppress.StopPredictionStatusTest do
       handle_events(@entries, :from, :state)
 
       assert MapSet.new([1, 2]) == flagged_stops_on_route("Red", 0)
+      assert MapSet.new([3, 4]) == flagged_stops_on_route("Red", 1)
+      assert MapSet.new([5]) == flagged_stops_on_route("Blue", 0)
+      assert MapSet.new([6]) == flagged_stops_on_route("Blue", 1)
     end
 
     test "returns nil if missing stop_id or direction_id" do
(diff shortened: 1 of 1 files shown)


config/config.exs
```diff
@@ -74,8 +77,8 @@ config :concentrate,
     Concentrate.GroupFilter.VehicleStopMatch,
     Concentrate.GroupFilter.SkippedStopOnAddedTrip,
     Concentrate.GroupFilter.TripDescriptorTimestamp,
-    Concentrate.GroupFilter.UncertaintyValue
-    # Concentrate.GroupFilter.SuppressStopTimeUpdate Enable once fully implemented
```
nlwstein, line 78: nit: I'm ambivalent personally, but there seems to be a preference here towards avoiding commented code. Maybe this should just be whacked, and then added when it's actually in use.

lib/concentrate/filter/suppress/supervisor.ex
```diff
@@ -0,0 +1,43 @@
...
+  @one_day 86_400_000
+
+  require Logger
+
+  def start_link(config) do
+    if config[:url] do
+      Supervisor.start_link(
+        [
+          {
+            Concentrate.producer_for_url(config[:url]),
```
nlwstein, line 17: Would it be worth adding a test to either confirm that it actually starts pulling this file?

nlwstein (approved):
Overall looks good, just a few minor thoughts that are non-blocking.

nlwstein (approved):

PR #359, opened 2024-09-16
update cancelled trips group filter to only cancel a trip if all stus are skipped for bus/route_type 3

bfauble:
#### Summary of changes

**Asana Ticket:** [🐛 [Investigate] Subway passthrough trips in Concentrate have schedule_relationship CANCELED](https://app.asana.com/0/584764604969369/1208163180298576/f)

* store route type for trips along with route_id and direction_id
* check route_type before applying `:CANCELED` status to trip when all stop time updates are `:SKIPPED`

### DIFF
--- a/test/concentrate/group_filter/cancelled_trip_test.exs
+++ b/test/concentrate/group_filter/cancelled_trip_test.exs
@@ -69,6 +69,7 @@ defmodule Concentrate.GroupFilter.CancelledTripTest do
     test "cancels the group if all stop updates have already been given a skipped status" do
       td =
         TripDescriptor.new(
+          route_id: "1",
           trip_id: "trip",
           start_date: {1970, 1, 2}
         )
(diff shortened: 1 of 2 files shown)


Whoops:
Checking my understanding, passthrough trips are nonrevenue trips where a train will pass through a station but won't pick up passengers, right? And they are currently being represented by a trip with all skipped stops?

So the idea here is that bus does not have these passthrough trips, so if we have all skipped stops it's a block waiver and we should cancel them, but subway it could be a passthrough?

I'm wondering if this is a safe assumption (that this logic should only every apply to bus, not Subway). Would it be possible to check the revenue status of the trip instead, or can we not rely on it being marked nonrevenue?

Whoops (approved):

Whoops:
Not seeing any glaring issues in dev-green, I think this is good to go.

Whoops:
I re-enabled CI (thanks Github for auto-disabling, SUPER helpful). I don't seem to be able to kick it into reruning for this though, I think you might have to re-push?

Whoops (approved):
I think the code in this version is easier to follow than the original, and it's looking good after running in dev blue for awhile 👍

PR #364, opened 2024-12-30
feat: cancel trips with a boarding status of Canceled

lemald:
#### Summary of changes

**Asana Ticket:** [[Investigate] Unexpected predictions during Commuter Rail cancellation](https://app.asana.com/0/584764604969369/1207967840855087/f)

I still need to do a test in dev with actual data, but this should be code-complete and ready to review.

I disabled the Credo cyclomatic complexity warning because I feel that the `filter/3` function is actually quite simple and readable despite the high complexity value, and since I don't see any obvious way to refactor it to be less "complex" per this metric that doesn't make it harder to follow.

### DIFF
--- a/lib/concentrate/group_filter/cancelled_trip.ex
+++ b/lib/concentrate/group_filter/cancelled_trip.ex
@@ -10,7 +10,6 @@ defmodule Concentrate.GroupFilter.CancelledTrip do
   @impl Concentrate.GroupFilter
   def filter(trip_group, module \\ CancelledTrips, routes_module \\ Routes)
 
-  # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
   def filter(
         {%TripDescriptor{} = td, _vps, [stu | _] = stop_time_updates} = group,
         module,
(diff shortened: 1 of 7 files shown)


Whoops (approved):

lemald:
My assumption going into this was that the StopTimeUpdates with the boarding status were going to be the only STUs for that trip, so we could do an `Enum.all?/2` very similar to how things work on bus with block waivers. However, some testing with some other changes reminded me that that might not actually be the case, so the schedule relationship at North Station wasn't updated as we would have expected. I'll have to revisit this and come back with a slightly different approach.

lemald:
@Whoops let me know what you think of the updated approach!

Whoops (approved):
Sorry, lost track of this! Looks good!

lemald:
@Whoops sorry, re-requesting your review yet again. I realized it's probably best if we use the same `skip/1` logic as other places in the application do, in particular for removing arrival and departure times. This necessitated a slight change to the typespec of `skip/1` because it was complaining about a mismatch between the opaque type `t` and the structured type - I think it's fine, but let me know what you thin.

Whoops (approved):

PR #367, opened 2025-05-14
feat: Screenplay API-based Prediction Suppressions

cmaddox5:
#### Summary of changes

**Asana Ticket:** [🧠 Use new Screenplay API to suppress predictions](https://app.asana.com/1/15492006741476/project/584764604969369/task/1209369662784554?focus=true)

Added a Screenplay API-based prediction suppression mechanism.


lib/concentrate/filter/suppress/rts/stop_prediction_status.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Concentrate.Filter.Suppress.StopPredictionStatus do
+defmodule Concentrate.Filter.Suppress.RTS.StopPredictionStatus do
```
rudiejd, line 1: suggestion: could this be RealTimeSigns instead of RTR?

lib/concentrate/filter/suppress/screenplay/stop_prediction_status.ex
```diff
@@ -0,0 +1,74 @@
...
+
+  defp log_unsuppressed_stops({_unchanged_entries, changed_entries}) do
+    Enum.each(changed_entries, fn %{
+                                    route_id: route_id,
+                                    direction_id: direction_id,
+                                    stop_id: stop_id,
+                                    suppression_type: suppression_type
+                                  } ->
+      Logger.info(
+        "Cleared prediction suppression for stop_id=#{stop_id} route_id=#{route_id} direction_id=#{direction_id} suppression_type=#{suppression_type} based on Screenplay API"
```
rudiejd, line 50: should this have an `event=` tag to be consistent with other logging for splunk summarization?

lib/concentrate/group_filter/screenplay_suppress_stop_time_update.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Concentrate.GroupFilter.ScreenplaySuppressStopTimeUpdate do
```
rudiejd, line 1: do we actually need a separate module, or could we just have a feature flag to switch from RTS to screenplay?

cmaddox5: The go-live plan for this is to add the new Screenplay logic and remove the old RTS logic. So the "final product" can be seen [here](https://github.com/mbta/concentrate/pull/368). No actual logic changes made in that PR, just stripped out all the RTS logic.

cmaddox5:
Just merged in the removal PR so we can see what the final version will be.

PR #370, opened 2025-06-04
feat(cancelled-trips): create SKIPPED StopTimeUpdates for cancelled trips

rudiejd:
#### Summary of changes

**Asana Ticket:** [🧠🐛 Fix up how RTR via Concentrate publishes canceled trips](https://app.asana.com/1/15492006741476/project/584764604969369/task/1210355969075349?focus=true)

Problem:
Concentrate sometimes receives trips with a `schedule_relationship` of `CANCELED` and no `StopTimeUpdate`s. As a result, first-party consumers such as MBTA Go and MBTA.com cannot update stop-level information for stops that a trip will skip.

Solution:
Add `SKIPPED` `StopTimeUpdate`s for every stop in a trip in RTR. I figured out which stops were in a trip by writing a new ETS query in the `GTFS.StopTimes` GenServer.


lib/concentrate/gtfs/stop_times.ex
```diff
@@ -62,6 +62,36 @@ defmodule Concentrate.GTFS.StopTimes do
...
+
+  defp select_stops_for_trip(trip_id) do
+    :ets.select(Concentrate.GTFS.StopTimes, [
+      {
+        # match specs are a cursed daemonic tongue
+        {{trip_id, :"$1"}, {:"$2", :_, :_, :_, :_, :_}},
+        [],
+        [:"$$"]
+      }
+    ])
```
lemald, line 90: **suggestion:** [this documentation on `select`](https://www.erlang.org/docs/23/man/ets#select-2) seems to imply that you could simplify this slightly with a call to `:ets.match` instead:

> The return value is constructed using the "match variables" bound in MatchHead or using the special match variables '$_' (the whole matching object) and '$$' (all match variables in a list), so that the following match/2 expression:

> `ets:match(Tab,{'$1','$2','$3'})`

> is exactly equivalent to:

> `ets:select(Tab,[{{'$1','$2','$3'},[],['$$']}])`

test/concentrate/group_filter/cancelled_trip_test.exs
```diff
@@ -83,12 +90,65 @@ defmodule Concentrate.GroupFilter.CancelledTripTest do
...
+          route_id: "1",
+          trip_id: "unknown",
+          start_date: {1970, 1, 2},
+          schedule_relationship: :CANCELED
+        )
+
+      group = {td, [], []}
+
+      {td_actual, [], []} =
+        filter(group, @module, @fake_routes_module, @fake_stop_times_module)
```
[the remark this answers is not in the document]
rudiejd: b2418fc

lib/concentrate/gtfs/stop_times.ex
rudiejd: b2418fc

PR #374, opened 2025-06-25
fix(cancelled-trips): re-implement making skipped stop time updates for cancelled trips

rudiejd:
#### Summary of changes

**Asana Ticket:** [🧠 Re-implement SKIPPED StopTimeUpdates for CANCELLED trips](https://app.asana.com/1/15492006741476/project/584764604969369/task/1210575419129897?focus=true)

Problem:
The PR that we previously deployed to add SKIPPED StopTimeUpdates for CANCELLED trips resulted in stop information "flickering" for the buses such that the stop to which a bus was headed would appear and disappear intermittently.

Solution:
This was happening because the `bus_block_waiver?/2` function returned `true` for trips with no `StopTimeUpdates`, and RTR intermittently gets bus trips with no `StopTimeUpdates` (likely from Busloc?). Previously, these trips were most likely ignored, but now they cause us to cancel bus trips unnecessarily, which causes us to [remove](https://github.com/mbta/concentrate/blob/647866e0734fd5ae36dda440631e57a69561c375/lib/concentrate/group_filter/vehicle_at_skipped_stop.ex#L1) the `stop_id` for the `VehiclePosition` entities. This PR makes it so that we only consider bus trips with all `SKIPPED` `StopTimeUpdates` as block waivers.

### DIFF
--- a/.envrc.example
+++ b/.envrc.example
@@ -1,0 +1,134 @@
+export CONCENTRATE_JSON=`cat <<EOF
+{
+  "alerts": {
+    "url": "https://mbta-realtime-test.s3.amazonaws.com/alerts_enhanced.json"
+  },
+  "gtfs": {
+    "url": "https://mbta-gtfs-s3.s3.amazonaws.com/MBTA_GTFS-dev-green.zip"
+  },
+  "signs_stops_config": {
+    "url": "s3://mbta-signs-dev/stops-config.json"
+  },
+  "log_level": "debug",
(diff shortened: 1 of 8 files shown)


rudiejd:
the commit to actually review here is https://github.com/mbta/concentrate/pull/374/commits/8a887ba1c23398a87fca65f2977b57532963c0ea - the other one is a revert of the revert 😅

PR #440, opened 2026-07-09
Better handle shuttle alert directionality

lemald:
#### Summary of changes

**Asana Ticket:** [🧠🐛Trains that run outside the limits of a diversion shouldn't have stops skipped](https://app.asana.com/1/15492006741476/project/584764604969369/task/1216121008812240?focus=true)

Inbound Lowell Line trains were originating at `NHRML-0218-01`, the normal outbound platform at North Billerica. Because the `informed_entity` in the alert (1017626) had activities of `BOARD` and `RIDE`, it was interpreted as the start of a shuttle, and therefore Concentrate skipped it and all of the stop time updates that came after it in the trip (since there was no other stop that counted as the end of the shuttle). However, the `informed_entity` also had a `direction_id` of 0, indicating that it was only effective in the outbound direction, and Concentrate wasn't respecting this. This change makes Concentrate reference the direction ID in the `informed_entity`, if present, to determine what kind of effect the alert is having on a given route and stop. If the direction ID is absent, it is assumed to apply in both directions.

Unfortunately this may be a bit hard to test without another very similar shuttle alert, schedule, and realtime tracking information. I'm hoping that the unit tests are sufficient and adequately replicate the scenario that caused the Lowell Line issue.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -35,7 +35,7 @@ jobs:
           key: ${{ runner.os }}-asdf-v2-${{ hashFiles('.tool-versions') }}
         id: asdf-cache
       # only run `asdf install` if we didn't hit the cache
-      - uses: asdf-vm/actions/install@v1
+      - uses: asdf-vm/actions/install@v4
         if: steps.asdf-cache.outputs.cache-hit != 'true'
       # if we did hit the cache, set up the environment
       - name: Setup ASDF environment
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -32,7 +32,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/.asdf
-          key: ${{ runner.os }}-asdf-v2-${{ hashFiles('.tool-versions') }}
+          key: ${{ runner.os }}-asdf-v4-${{ hashFiles('.tool-versions') }}
         id: asdf-cache
       # only run `asdf install` if we didn't hit the cache
       - uses: asdf-vm/actions/install@v4
--- a/lib/concentrate/filter/alert/shuttles.ex
+++ b/lib/concentrate/filter/alert/shuttles.ex
@@ -39,9 +39,13 @@ defmodule Concentrate.Filter.Alert.Shuttles do
           direction_id :: 0 | 1,
           :calendar.date() | integer
         ) :: :start | :stop | :through | nil
-  def stop_shuttling_on_route(route_id, stop_id, _direction_id, date_or_timestamp)
+  def stop_shuttling_on_route(route_id, stop_id, direction_id, date_or_timestamp)
       when is_binary(route_id) and is_binary(stop_id) do
-    case TimeTable.date_overlaps(@table, {:route_stop, route_id, stop_id}, date_or_timestamp) do
+    case TimeTable.date_overlaps(
+           @table,
+           {:route_stop, route_id, stop_id, direction_id},
+           date_or_timestamp
+         ) do
       [atom | _] -> atom
       [] -> nil
     end
@@ -81,9 +85,10 @@ defmodule Concentrate.Filter.Alert.Shuttles do
 
     route_stops =
       if is_binary(stop_id) and is_binary(route_id) do
-        [
-          {{:route_stop, route_id, stop_id}, shuttle_type(InformedEntity.activities(entity))}
-        ]
+        for direction_id <- direction_ids(InformedEntity.direction_id(entity)) do
+          {{:route_stop, route_id, stop_id, direction_id},
+           shuttle_type(InformedEntity.activities(entity))}
+        end
       else
         []
       end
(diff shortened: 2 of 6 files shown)


.github/workflows/elixir.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           key: ${{ runner.os }}-asdf-v2-${{ hashFiles('.tool-versions') }}
         id: asdf-cache
       # only run `asdf install` if we didn't hit the cache
-      - uses: asdf-vm/actions/install@v1
+      - uses: asdf-vm/actions/install@v4
```
jzimbel-mbta, line 38: suggestion:
Should the cache key also be updated to avoid using a cached install from the old version of asdf / the asdf install action? Like change `-v2-` to `-v4-` in the key, perhaps.

lib/concentrate/filter/alert/shuttles.ex
```diff
@@ -80,9 +85,10 @@ defmodule Concentrate.Filter.Alert.Shuttles do
 
     route_stops =
       if is_binary(stop_id) and is_binary(route_id) do
-        [
-          {{:route_stop, route_id, stop_id}, shuttle_type(InformedEntity.activities(entity))}
-        ]
+        for direction_id <- direction_ids(InformedEntity.direction_id(entity)) do
+          {{:route_stop, route_id, stop_id, direction_id},
+           shuttle_type(InformedEntity.activities(entity))}
```
jzimbel-mbta: suggestion:
Since the value will be the same for all elements of the list, you could evaluate `shuttle_type(InformedEntity.activities(entity))` once outside of this comprehension.

(Bit of a micro-optimization tbh, since the list has only 3 elements max)

lib/concentrate/group_filter/shuttle.ex
```diff
@@ -31,27 +31,27 @@ defmodule Concentrate.GroupFilter.Shuttle do
 
   def filter(%TripGroup{} = other, _module), do: other
 
-  defp shuttle_updates(route_id, stus, module) do
+  defp shuttle_updates(route_id, direction_id, stus, module) do
     initial_state = {false, false}
```
jzimbel-mbta: comment:
Not part of the changes here, but it's difficult to determine what exactly this pair of booleans represents since the first time they're labeled in any way is in the clause on line 48. It would be helpful to readers if we did something like this:
```suggestion
    initial_state = {_has_started? = false, _has_shuttled? = false}
```
or used a map instead.

lib/concentrate/filter/alert/shuttles.ex
```diff
@@ -36,11 +36,16 @@ defmodule Concentrate.Filter.Alert.Shuttles do
...
+          direction_id :: 0 | 1,
           :calendar.date() | integer
         ) :: :start | :stop | :through | nil
-  def stop_shuttling_on_route(route_id, stop_id, date_or_timestamp)
+  def stop_shuttling_on_route(route_id, stop_id, direction_id, date_or_timestamp)
       when is_binary(route_id) and is_binary(stop_id) do
-    case TimeTable.date_overlaps(@table, {:route_stop, route_id, stop_id}, date_or_timestamp) do
+    case TimeTable.date_overlaps(
+           @table,
+           {:route_stop, route_id, stop_id, direction_id},
```
jzimbel-mbta, line 46: question:
Does the shuttles timetable ETS table contain both 3-tuples that do not include direction ID, and 4-tuples that do include direction ID?

I guess I'm a little surprised that you didn't need to make any corresponding changes over in the TimeTable module (or [here?](https://github.com/mbta/concentrate/blob/d5166cccf7ccc27966c6cad2035c6576b3469728/lib/concentrate/filter/alert/shuttles.ex#L62-L70)) for this new record structure.

(I admit I'm a little rusty on ETS MatchSpecs, so I might be misunderstanding how this key ends up getting used in the `select` query.)

jzimbel-mbta (approved):
This looks good overall, just some minor suggestions and a sanity check about the updated select query.

My understanding of the approach is that direction ID (when present in an informed entity) is now included in the TimeTable ETS select query so that we avoid selecting shuttle data that doesn't apply to a trip in this direction. Does that sound right?

lib/concentrate/filter/alert/shuttles.ex
lemald: They now contain only the four-tuples - I didn't directly change the code in your "here" link, but I did change the `cancellation_type/1` function that it calls, which is how we generate the tuples that get added to the table.

lemald:
> My understanding of the approach is that direction ID (when present in an informed entity) is now included in the TimeTable ETS select query so that we avoid selecting shuttle data that doesn't apply to a trip in this direction. Does that sound right?

Correct!

.github/workflows/elixir.yml
lemald: 4d1b9be

lib/concentrate/filter/alert/shuttles.ex
lemald: c6e841f

PR #442, opened 2026-07-14
Add zizmor to concentrate

npatel-mbta:
#### Summary of changes

**Asana Ticket:** [Add Zizmor to concentrate](https://app.asana.com/1/15492006741476/project/584764604969369/task/1215634628813616?focus=true)

[Please include a brief description of what was changed]


.github/workflows/docker.yml
```diff
@@ -1,11 +1,19 @@
 name: Docker
 
+permissions: {}
```
jzimbel-mbta: suggestion:
I think they're not a hard requirement because the project is public, but I'd still recommend adding `contents: read` here in case that changes for some reason in the future. That permission is needed by `actions/checkout` in private repos.

.github/workflows/elixir.yml
```diff
@@ -70,12 +79,30 @@ jobs:
...
           name: elixir-lcov
           path: coverage/
-      - uses: mbta/actions/dialyzer@v2
+      - uses: mbta/actions/dialyzer@d37877022d4ae12e96663d2e0a970a5214a2e232 # v2.20
+
+  zizmor:
+    name: Run zizmor
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read # required for checkout
```
jzimbel-mbta: suggestion:
Should be safe to remove this since `contents: read` is already given at the workflow scope.

.github/workflows/report_coverage.yml
```diff
@@ -1,53 +1,68 @@
 name: Report coverage
 
-on:
+on: # zizmor: ignore[dangerous-triggers]
```
jzimbel-mbta: suggestion:
If you'd like an example to follow for addressing this zizmor complaint, see [this commit](https://github.com/mbta/ocs_saver/pull/523/changes/94ae696e40660059470b98f1d7290e1b789aa086) in my ocs_saver PR. It's pretty straightforward overall.

_Or_... you could simply delete this entire workflow file since [it has never run successfully](https://github.com/mbta/concentrate/actions/workflows/report_coverage.yml?query=is%3Asuccess). 🚮

jzimbel-mbta (approved):
3 non-blocking suggestions, but looks good overall.

Please run a deploy to one or more of the dev-color environments before merging, to confirm that they work.

.github/workflows/docker.yml
npatel-mbta: added, thank you

.github/workflows/elixir.yml
npatel-mbta: removed

.github/workflows/report_coverage.yml
npatel-mbta: removed workflow file

npatel-mbta:
Deployed to dev-blue

PR #445, opened 2026-07-29
feat(ci): require tagged releases for deploy to prod

lemald:
#### Summary of changes

**Asana Ticket:** [Set up tag/release-based prod deploys - concentrate](https://app.asana.com/1/15492006741476/project/584764604969369/task/1216046041063093?focus=true)

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -1,0 +1,37 @@
+name: Deploy to prod
+
+permissions: {}
+
+on:
+  workflow_dispatch:
+  release:
+    types:
+      - released
+
+concurrency:
+  group: deploy-prod
+  cancel-in-progress: true
+
+jobs:
+  check-ref:
+    name: Fail if not a tag
+    if: ${{ !startsWith(github.ref, 'refs/tags/') }}
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,37 @@
...
+    types:
+      - released
+
+concurrency:
+  group: deploy-prod
+  cancel-in-progress: true
+
+jobs:
+  check-ref:
+    name: Fail if not a tag
```
huangh, line 17: so this just checks if it's a tag - and not necessarily a release. by convention i assume we only use tags for releases? but any tag could be released here. is there a tagging convention that we should check for and enforce here?

huangh (approved):
I think it looks fine as is - question as to whether we want to lock down the deploy-prod further, but i think the intent of this all is to give us a bit more flexibility so i'm ok with not addressing my comment

huangh:
oh...we have no tags. hmm...as a part of this, are we enforcing a convention for tagging then? v1.0.0 etc?

huangh:
after discussion in asana, seems like the tooling isn't fully sufficient or worth the effort to do what i'm describing. ok with proceeding as is

PR #449, opened 2026-09-11
feat(GroupFilter): Propogate CR Delays to downstream stops

KaylaBrady:
#### Summary of changes

**Asana Ticket:** [CR Delays | Concentrate change to propagate status when delays](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1218067460223114?focus=true)

This adds a new GroupFilter to propagate delay StopTimeUpdates under certain circumstances:
* CR trips only
* There is a status-only StopTimeUpdate indicating a delay
* For downstream stops that have scheduled departure in the past and no existing StopTimeUpdate

So far I have only unit tested this and am not sure how to best test it end-to-end (are we able to mock status-only delay STUs coming from Keolis?)

Note: I used AI for autocompleting helper functions and generating test stubs.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -76,7 +76,9 @@ config :concentrate,
     Concentrate.GroupFilter.SkippedStopOnAddedTrip,
     Concentrate.GroupFilter.TripDescriptorTimestamp,
     Concentrate.GroupFilter.UncertaintyValue,
-    {Concentrate.GroupFilter.SuppressStopTimeUpdate, terminal_suppression_by_time: %{}}
+    {Concentrate.GroupFilter.SuppressStopTimeUpdate, terminal_suppression_by_time: %{}},
+    {Concentrate.GroupFilter.PropogateDelayedStatus,
+     first_stop_delaying_statuses: ["Delayed"], downstream_status: "Delayed"}
   ],
   source_reporters: [
     Concentrate.SourceReporter.Basic,
--- a/config/config.exs
+++ b/config/config.exs
@@ -77,7 +77,7 @@ config :concentrate,
     Concentrate.GroupFilter.TripDescriptorTimestamp,
     Concentrate.GroupFilter.UncertaintyValue,
     {Concentrate.GroupFilter.SuppressStopTimeUpdate, terminal_suppression_by_time: %{}},
-    {Concentrate.GroupFilter.PropogateDelayedStatus,
+    {Concentrate.GroupFilter.PropogateDownstreamDelays,
      first_stop_delaying_statuses: ["Delayed"], downstream_status: "Delayed"}
   ],
   source_reporters: [
--- a/config/config.exs
+++ b/config/config.exs
@@ -78,7 +78,7 @@ config :concentrate,
     Concentrate.GroupFilter.UncertaintyValue,
     {Concentrate.GroupFilter.SuppressStopTimeUpdate, terminal_suppression_by_time: %{}},
     {Concentrate.GroupFilter.PropogateDownstreamDelays,
-     first_stop_delaying_statuses: ["Delayed"], downstream_status: "Delayed"}
+     matching_statuses: ["Delayed"], downstream_status: "Delayed"}
   ],
   source_reporters: [
     Concentrate.SourceReporter.Basic,
--- a/config/test.exs
+++ b/config/test.exs
@@ -21,7 +21,7 @@ config :concentrate, :group_filters, [
      }
    }},
   {Concentrate.GroupFilter.PropogateDownstreamDelays,
-   first_stop_delaying_statuses: ["Delayed"], downstream_status: "Delayed"}
+   matching_statuses: ["Delayed"], downstream_status: "Delayed"}
 ]
 
 config :concentrate, :sink_s3, ex_aws: Concentrate.TestExAws
--- a/config/test.exs
+++ b/config/test.exs
@@ -19,7 +19,9 @@ config :concentrate, :group_filters, [
        6 => {~T[04:00:00], ~T[07:30:00]},
        7 => {~T[04:00:00], ~T[07:30:00]}
      }
-   }}
+   }},
+  {Concentrate.GroupFilter.PropogateDownstreamDelays,
+   first_stop_delaying_statuses: ["Delayed"], downstream_status: "Delayed"}
 ]
 
 config :concentrate, :sink_s3, ex_aws: Concentrate.TestExAws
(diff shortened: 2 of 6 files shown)


config/test.exs
```diff
@@ -19,7 +19,9 @@ config :concentrate, :group_filters, [
        6 => {~T[04:00:00], ~T[07:30:00]},
        7 => {~T[04:00:00], ~T[07:30:00]}
      }
-   }}
+   }},
+  {Concentrate.GroupFilter.PropogateDownstreamDelays,
+   first_stop_delaying_statuses: ["Delayed"], downstream_status: "Delayed"}
```
KaylaBrady: TODO: Get complete list of relevant statuses. I don't love the name `first_stop_delaying_statuses` either. Maybe just `matching_statuses` would be sufficient

lib/concentrate/group_filter/propagate_downstream_delay_statuses.ex
```diff
@@ -0,0 +1,106 @@
...
+
+    Enum.flat_map(scheduled_stop_times, fn scheduled_stop_time ->
+      {stop_sequence, stop_id, _arrival, departure} = scheduled_stop_time
+      existing_stu = Map.get(stop_sequence_to_stu, stop_sequence)
+
+      cond do
+        existing_stu != nil ->
+          [existing_stu]
+
+        !is_nil(departure) && now > departure && stop_sequence > first_stop_sequence ->
```
KaylaBrady, line 84: note: I don't compare stop_ids anywhere and am only relying on stop_sequence. Is that safe?

config/config.exs
```diff
@@ -76,7 +76,9 @@ config :concentrate,
     Concentrate.GroupFilter.SkippedStopOnAddedTrip,
     Concentrate.GroupFilter.TripDescriptorTimestamp,
     Concentrate.GroupFilter.UncertaintyValue,
-    {Concentrate.GroupFilter.SuppressStopTimeUpdate, terminal_suppression_by_time: %{}}
+    {Concentrate.GroupFilter.SuppressStopTimeUpdate, terminal_suppression_by_time: %{}},
+    {Concentrate.GroupFilter.PropogateDownstreamDelays,
+     matching_statuses: ["Delayed"], downstream_status: "Delayed"}
```
KaylaBrady, line 81: TODO: Get complete list of relevant statuses.

lib/concentrate/group_filter/propagate_downstream_delay_statuses.ex
lemald, line 1: **nitpick / suggestion:** I think a name along the lines of `PropogateDownstreamDelayStatuses` might be slightly more descriptive and make it clear that it's about the boarding status field, not the actual predicted stop times.

KaylaBrady: Thank you, that is clearer!

test/concentrate/group_filter/propogate_downstream_delay_statuses_test.exs
```diff
@@ -0,0 +1,192 @@
...
+          trip_id: "trip",
+          start_date: {2026, 1, 1},
+          schedule_relationship: :CANCELED
+        )
+
+      stu = StopTimeUpdate.new(trip_id: "trip", status: :CANCELED)
+
+      group = %TripGroup{td: td, stus: [stu]}
+
+      assert group == PropogateDownstreamDelayStatuses.filter(group)
```
lemald: **question:** should this test also pass in a mock `now_fn` like the next test does, to make sure that the extra STUs would have been added had the `schedule_relationship` been different?

lib/concentrate/group_filter/propagate_downstream_delay_statuses.ex
lemald: I think it's fine! The combination of trip ID (these are all from the same trip) and stop sequence should uniquely identify the schedule departure.

test/concentrate/group_filter/propogate_downstream_delay_statuses_test.exs
KaylaBrady: TY, updated the test and confirmed it fails now if I change the schedule_relationship

lib/concentrate/group_filter/propagate_downstream_delay_statuses.ex
KaylaBrady: Great, I have a vague memory of something about virtual vs physical platforms for CR which I why I was hoping to avoid it in the first place.

lemald: Ah, right, I had forgotten about the parent vs. child stop thing entirely but that is indeed a good reason for relying just on stop sequence.

lemald (approved):
Okay, I think this looks good! It sounds like before merging and going to prod we'll want to:

1. Confirm the list of statuses for which this should apply (already noted).
2. Deploy to a dev environment and test out there.

config/config.exs
KaylaBrady: @lemald I updated the status list to re-use the config from boarding_status_override based on the [given list.](https://mbta.slack.com/archives/CT1LX7Y1L/p1789416351681679?thread_ts=1787944328.535289&cid=CT1LX7Y1L) My one concern is that both "PRIORITY" and "HOLD" Statuses are converted to "Info to follow". Any concerns about this logic applying to "PRIORITY" too? 

Since the conversion of boarding_status_override is done at the parsing level I think it would be a bigger but not horrible effort to change (maybe just a new filer at the end). But if the user-facing string is the same, I'm wondering if it would be acceptable to treat them as the same for the purpose of downstream delays.

lemald: Yeah, I think that's probably fine, especially since I'm guessing that `PRIORITY` and `HOLD` are less-frequently-used statuses.

lib/concentrate/group_filter/propagate_downstream_delay_statuses.ex
```diff
@@ -0,0 +1,114 @@
...
+        routes_module \\ Routes,
+        now_fn \\ &now/0
+      ) do
+    route_id = TripDescriptor.route_id(td)
+
+    if routes_module.route_type(route_id) != 2 ||
+         TripDescriptor.schedule_relationship(td) != :SCHEDULED do
+      group
+    else
+      Logger.info("#{__MODULE__} processing trip_id=#{inspect(TripDescriptor.trip_id(td))} inspected=#{inspect(td)}")
```
KaylaBrady: Logging to help catch an issue where the trip_date passed to `stops_for_trip_with_arrival_departure` was nil

